PR #93, opened 2016-04-06
Update README.md

TomKita:
Added clarification that devtools were disabled in production build in vue 1.0.18+


yyx990803:
Thanks!

Issue #101, opened 2016-05-16
Feature idea: export / import Vuex store's state


zigomir:
## Use case

> Your QA team is testing an app and they find an issue at some point. They open `vue-devtools` and under `vuex` tab there is export button that stringifies whole `store.state`. With this button, whoever finds a bug can easily copy and paste whole state and share it with developer(s). Developer then opens same app on local dev environment, opens up `vue-devtools` where under `Vuex` tab there is `input` or `textarea` where store's state can be pasted in. `vue-devtools` just parse state from input and populate real store's state with it. **Bug is immediately reproduced.**

I have a branch where I'm playing with this approach. Is this something we want to have in dev tools? 
It's actually not a lot of work to just make it work, but I haven't done any proper styling etc. Before going on with more work I wanted to sync up first.

This is idea is not mine, I first heard it from [Richard Feldman](https://github.com/rtfeldman) of Elm fame.

zigomir:
Done by #102

PR #102, opened 2016-05-17
Export/Import state

zigomir:
Here is implementation of idea from [issue 101](https://github.com/vuejs/vue-devtools/issues/101).
It's still a bit rough, but I think it can be super useful. I still need to try it out on larger objects.

vuex dev tools export import state

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "babel-loader": "^6.0.0",
     "babel-plugin-transform-runtime": "^6.3.13",
     "babel-preset-es2015": "^6.3.13",
-    "babel-runtime": "^5.8.0",
+    "babel-runtime": "^6.6.1",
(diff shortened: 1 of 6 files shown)


yyx990803:
I think the import should also be a button, and have a popup textarea only when clicked.

zigomir:
@yyx990803 I updated code and gif to use popup. Not sure if we want any close buttons or anything. It now closes on ESC key or clicking on `paste` button again.

yyx990803:
niceeeee

yyx990803:
Great job - I will add a few styling tweaks before doing a release.

yyx990803:
btw would be nice to add a few assertions in the test: https://github.com/vuejs/vue-devtools/blob/master/test/specs/test.js

Issue #107, opened 2016-05-26
Enhancement: sort keys alphabetically


simplesmiler:
When displaying objects in console, Chrome sorts the keys alphabetically. This makes it easy to find the key you are looking for.

![chrome-key-ordering](https://cloud.githubusercontent.com/assets/550229/15577588/d3b4d734-2365-11e6-8b59-275244fc7a61.png)

PR #108, opened 2016-05-26
sorting keys alphabetically

bartlomieju:
ref #107 

Now keys in inspector are displayed alphabetically, additionally I removed `limitBy` so it should work with Vue 2.0


yyx990803:
Awesome, thanks!

PR #148, opened 2016-09-28
Fix sorting of component properties in component inspector

markussorg:
The component properties within the component inspector are not always sorted in the correct alphabetical order. An example of this can be found in the TransitionGroup component in the "HackerNews Clone Example" from the docs:

unsorted_properties

This pr fixes the sorting of the properties.


skyrpex:
Didn't notice that... Maybe it would be worth to allow different sorting methods: Alphabetical, attribute type (data, prop, computed), ...

Issue #136, opened 2016-08-17
Display all args of events in Event tab


miljan-aleksic:
Now that Vue 2 events scope is limited it would be relatively easy to add a section Events for each component where those would be logged showing the arguments as well. I think that would be useful for debuging and knowing more about what is going on under the hood.

frankdugan3:
> showing the _**arguments**_

We now have the tab thanks to #159, but it only shows the first arg as the payload. Would be great to see all the args of the event.

yyx990803:
@frankdugan3 FYI the payload is an array representing the arguments.

PR #159, opened 2016-10-22
Add event inspection tab

markussorg:
This PR adds an event tab to inspect custom events on components and the data which is sent along with it. References #136. Works with Vue 1.x and 2.x.
## Implementation details

To send events to vue-devtools, it hooks into the $vm._events object. I don't know if this is a valid approach.  
## Preview

vue_devtools_events


skyrpex:
The event list would be overpopulated in a matter of seconds... I wonder if it should allow the developer to skip some event types.

cfreear:
This is just what I need for my current project! @skyrpex makes a good point too, a filter would be invaluable for this.

yyx990803:
Very cool indeed, but yeah a filtering mechanism would be great. (Vuex pane needs that too)

yyx990803:
Hi @markussorg , it would be nice to have the same filtering option like implemented in https://github.com/vuejs/vue-devtools/pull/172

If you don't have time to work on that right now, no worries. I can merge and add it later.

markussorg:
@yyx990803 I added search to the current pull-request:

search

src/backend/events.js
```diff
@@ -0,0 +1,28 @@
...
+export function initEventsBackend (bridge, instanceMap, getInstanceName) {
+  bridge.on('trigger-event', (event) => {
+    const instance = instanceMap.get(event.instanceId)
+    instance._events[event.eventName][0](event.eventData)
+  })
+  instanceMap.forEach((instance) => {
+    const listeners = instance._events
+    if (listeners) {
+      Object.keys(listeners).forEach((eventName) => {
+        if (!eventName.startsWith('hook:')) {
```
markussorg: @yyx990803 I don't know if this is a valid approach to hook into the vue event system. I don't know too much about the internals of vue that I can say if this can have any side effects or not.

yyx990803: This is fine, all internal events start with `hook:`.

src/backend/events.js
```diff
@@ -0,0 +1,28 @@
+import { stringify } from '../util'
+
+export function initEventsBackend (bridge, instanceMap, getInstanceName) {
```
yyx990803: After taking another look, I think there is a problem with this implementation: it only patches the listeners for the instances scanned when the backend inits, so it doesn't affect instances created later (e.g. by toggling a `v-if`).

A better way could be simply patching `Vue.prototype.$emit`.

markussorg:
@yyx990803 I updated this pull request. It works now with conditionally rendered components too.

src/backend/index.js
```diff
@@ -31,6 +32,7 @@ export function initBackend (_bridge) {
       initVuexBackend(hook, bridge)
     })
   }
+  initEventsBackend(rootInstances[0], bridge, instanceMap, getInstanceName)
```
markussorg: @yyx990803 I don't know if this is the right approach!? I grab the first root instance of the page and patch the prototype there.

yyx990803: You can access the global Vue via `hook.Vue` here

src/backend/index.js
```diff
@@ -31,6 +32,7 @@ export function initBackend (_bridge) {
       initVuexBackend(hook, bridge)
     })
   }
+  initEventsBackend(hook.Vue, bridge, instanceMap, getInstanceName)
```
markussorg: @yyx990803 I'm passing global Vue via hook.Vue now!

yyx990803:
Great work!

PR #168, opened 2016-11-06
Update conf for webpack 2.1.0-beta.25

fayway:
Moving vue-loader options to module.rules since webpack 2 no longer allows custom properties in configuration


yyx990803:
Thanks, this is addressed by #161 , and latest vue-loader doesn't require explicitly configuring buble loader.

Issue #134, opened 2016-08-06
Feature request for vuex tree


davidmoshal:
Couple of nice features 'just saying':

- filter_by_mutation_name(/regex/)
- exand-all(), close-all()
- remember expanded state after hot reload 
- flash node changes to first visible parent

thanks for the excellent project.

skyrpex:
> * filter_by_mutation_name(/regex/)
* exand-all(), close-all()
* remember expanded state after hot reload

I agree with all of them. :+1: 

> * flash node changes to first visible parent

What do you mean with that?

davidmoshal:
>  - flash node changes to first visible parent

if tree collapsed and a non-visible node changes, then indicate that an invisble node has changed by animating the visible parent.

By "flash" i mean't animate a color change of the parent.

skyrpex:
Oh, yeah. That's pretty useful too!

davidmoshal:
one last thing: often vue-devtool fails to start, which sometimes requires quitting and restarting Chrome - which is state breaking - I'm guessing this is a known issue, open elsewhere.

kaleb:
In other parts of the chrome dev tools, alt-clicking (ctrl+alt click for me on Unity desktop) performs an expand-all:

https://umaar.com/dev-tips/18-alt-click-expand-node/

Issue #112, opened 2016-05-31
Filtering


dddomin3:
Vue(x) Dev Tools should have a ability to filter both state, and mutation names. 
If possible, having the ability to categorize mutations with a string, enum, and/or a log level. Either through special dispatch attributes that are only considered/passed to dev middlewares, or whatever is a better implementation.

I feel as though the vuex dev tools quickly gets filled. When mutations involving many different interactions (API state, page state, component lifecycle state) all end up all in one flat list, it's hard to track an issue in a specific functionality.

stealthybox:
Maybe we should PR Vuex to modify the [Object-Style Dispatch](http://vuex.vuejs.org/en/mutations.html) to include a `category` parameter?

This is where we would start adding that to make it consumable by the devtools:
https://github.com/vuejs/vuex/blob/master/src/index.js#L84

Namek:
I would be happy to have a **simple text input** that would filter by action names.


My case:
I've implemented a small in-app calendar where some mouse down/up/enter/leave events are triggered and based on those events some calendar days are selected to be further processed. 

`mouseenter` is the worst event, because I have action dispatched for it so it messes action list in Vuex DevTools very quickly. Which obviously makes debugging a lot harder.


![image](https://cloud.githubusercontent.com/assets/31058/18029899/7f9d9f76-6ca5-11e6-98d0-f4cce66950f1.png)

stealthybox:
@Namek, as a temporary workaround, you can switch to the Object-Style Dispatch and use the [silent flag](http://vuex.vuejs.org/en/mutations.html) for your `MOUSE_ENTER` mutations.

Hopefully, we can all pool together to get this implemented soon.

posva:
Hey @dddomin3, sorry for the delay.
About filtering mutations, it's now possible. If you organise your store using modules, you can easily filter mutations of one module.
About filtering the state. I'm not sure to understand that from a UX perspective. Can you explain the case when you see yourself _filtering_ the state?

dddomin3:
Sure! Sorry for my late response. The mutation filtering improvements are great! Good job there :+1: 

As for state filtering:
Let's say I'm working on populating a certain piece of the state (like a user's detailed data). It'd be nice to be able to specify `user.details.favoriteColor,` and only display that in the store. Almost like `_.get()` from lodash:
https://lodash.com/docs/#get
Although, there may be a better way to UX that, that's just my imagining of it.

[...]

PR #172, opened 2016-11-11
adds mutation filtering to Vuex tab

bartlomieju:
I guess this feature has been requested several times already  (#134, #112 ), so went ahead and implemented it. It allows to use both simple string search and RegExes. Appropriate error is displayed if RegEx can't be compiled.

vue-devtools

Unfortunately I couldn't get tests to start, so I did not write any. I'll try to do that but any help would be appreciated.

### DIFF
--- a/src/devtools/components/VuexHistory.vue
+++ b/src/devtools/components/VuexHistory.vue
@@ -13,6 +13,14 @@
         <i class="material-icons">cached</i>
         <span>Reset</span>
       </a>
+      <div class="search">
+        <i class="material-icons">search</i>
+        <input class="search-box" placeholder="Filter mutations" v-model="userInputFilter">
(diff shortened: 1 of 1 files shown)


yyx990803:
Nice! I tested locally and it's working great. But ideally we still want tests. What errors are you getting from `npm test`?

bartlomieju:
Actually none... It hangs on 'Running:  vue-devtools e2e tests'. Sometimes it doesn't open browser, sometimes it does but with url 'data:;' and sometimes it opens actuall dev server but tests are not running. It deleted and re-installed whole package several times with no luck. Later I'll try to purge Java on my machine.

yyx990803:
hmm... ok, I'll merge and add tests later, thanks!

davidmoshal:
@bartlomieju thanks!

PR #173, opened 2016-11-13
Avoid using document.all

pifantastic:
`document.all` returns an `HTMLAllCollection` which has some, uhm, [special](http://stackoverflow.com/questions/10350142/why-is-document-all-falsy), behavior. This is causing errors when es6-shim is present, which in some cases overwrites the native `Array.prototype.find` with its own implementation.

This is an attempt to prevent vue-devtools from throwing an error in Chrome when es6-shim is present.

See: https://github.com/paulmillr/es6-shim/issues/428

### DIFF
--- a/shells/chrome/src/detector.js
+++ b/shells/chrome/src/detector.js
@@ -6,7 +6,8 @@ window.addEventListener('message', e => {
 
 function detect (win) {
   setTimeout(() => {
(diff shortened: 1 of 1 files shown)


ljharb:
It's also worth noting that `document.all` has been heavily deprecated for almost a decade now, and should never be used, regardless of library bugs with it.

PR #181, opened 2016-11-18
fix [].find error

xqin:
Chrome (`version: 44.0.2403.125`) does not support  `[].find`, so change to `[].filter`.

### DIFF
--- a/shells/chrome/src/detector.js
+++ b/shells/chrome/src/detector.js
@@ -7,7 +7,7 @@ window.addEventListener('message', e => {
 function detect (win) {
   setTimeout(() => {
     const all = document.querySelectorAll('*')
-    const el = [].find.call(all, e => e.__vue__)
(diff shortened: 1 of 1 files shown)


xqin:
Vue.js devtools version: `2.3.1`
image

yyx990803:
Why would you be using Chrome 44? It's 10 versions behind the current.

LinusBorg:
Coincidentally I had a discussion in an issue some days ago, with someone wroking in a bank that was "stuck" on v44 because of company policy. @simplesmiler suggested this could be because v44 was the last one to support the Netscape plugin AIP (NPAPI)

ChalkPE:
`find` stops after having found an element, while `filter` scans entire array even it already found one.

yyx990803:
Closing as this is addressed by #240

PR #185, opened 2016-11-20
Change attached and detached lifecycle hooks in key-nav mixin

dominiczaq:
Changed for mounted and destroyed respectively to fix not working arrow keys navigation.

### DIFF
--- a/src/devtools/mixins/key-nav.js
+++ b/src/devtools/mixins/key-nav.js
@@ -18,10 +18,10 @@ document.addEventListener('keyup', e => {
 })
 
 export default {
-  attached () {
+  mounted () {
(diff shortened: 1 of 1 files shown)


dominiczaq:
Hello,
I've always wondered why vue-devtools doesn't have support for arrow keys navigation in component tree view. I've decided to check it out in code and it looks like it was already there - the only problem was that it was still using removed lifecycle hooks (attached and detached). After changing to mounted and destroyed keys navigation works like charm.

Hope this could be merged and thank you for all the hard work you put into Vue.js!

yyx990803:
woah, totally missed this one upgrading to 2.0! thanks!

PR #186, opened 2016-11-22
Update Readme

reustonium:
Removed readme item specifying that Vue-Devtools only works with Vue v1.x.
These tools seem to work fine with Vue 2.x.


yyx990803:
1.0.0+ includes 2.0 ;) But yeah this can be removed since I doubt any still uses 0.x versions.

Issue #60, opened 2015-12-10
Scrollbar


ksdev-pl:
Hello

Could you please add scrollbars? With mouse without scroll or RSI (with which you should not use mouse scroll), it's not very comfortable to scroll long lists.

Thanks for your great work.

skyrpex:
Yes please. I came here for that reason.

kaleb:
I am seeing scrollbars in mine in the components and the events views. Where are you not seeing scrollbars where there should be scrollbars?

ksdev-pl:
It's an issue for old devtools version, should be closed.

PR #191, opened 2016-12-02
Add scrollbars

ksdev-pl:
Related to: #60.

Adds a scrollbar visible only when needed to the left and right panel in components tab.


ksdev-pl:
As #193 was merged this is no longer needed.

PR #193, opened 2016-12-05
consistency and responsive adjustments

crswll:
There's a bunch of updates here. I still want to clean up some code but it's getting there. I'll try to summarize a bit.

## ScrollPane
This component allows the top to stay put and the rest to scroll with overlay scrolling.

## Consistency
The search bar is now always to the left with buttons to the far right of it. They're all very squishy too.

## Responsive
We add more comfortable spacing and show icon text based on the width / height of the devtools. Looks "good" from 450px up.

## CSS Cleanup
The "active" color is set through one variable now, so basically everywhere you see blue can be easily switched to another color. I'd like to add a "switch" to enable a dark mode a little bit down the line as well.

Excited to hear what you all think.

### DIFF
--- a/shells/dev/index.html
+++ b/shells/dev/index.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf8">
     <title>Plain Shell</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <style>
       #target {
         flex: 1;
--- a/shells/dev/target/Other.vue
+++ b/shells/dev/target/Other.vue
@@ -31,7 +31,8 @@ export default {
           d: null,
           e: undefined,
           f: true,
-          g: 12345
+          g: 12345,
+          h: 'I am a really long string mostly just to see how the horizontal scrolling works.',
         }
       }
     }
--- a/src/devtools/common.styl
+++ b/src/devtools/common.styl
@@ -7,7 +7,7 @@ $darkerGreen = #3ba776
 $wide = 820px
 
 // The min-height to give the tools a little more breathing room...
-$tall = 500px
+$tall = 300px
 
 // Theme
 $active-color = $darkerGreen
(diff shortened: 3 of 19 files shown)


crswll:
Hey @yyx990803 

Not sure who gets notified or what on these comments/commits but I think this is good to go if you wanted to check it out and let me know what you think.

ksdev-pl:
Hi

I'm curious about "0 width horizontal scroll" - how would you scroll a long value and see (at first sight) how long the horizontal values are?

And the "minimal vertical scrollbar" - isn't the scrollbar is for clicking? Making the width minimal makes it harder (but it's great to have a scrollbar :).

Don't take it too serious, I'm feeling like an intruder talking about it (your work is great), I just don't understand the removal of scrollbars, especially in developer tools (and not some fancy hipster website where appearance is more important than usability) - this should be especially pointed at Evan, as we don't have any scrollbars at this moment ;)

crswll:
Hey,

I don't think you're an intruder. I appreciate the feedback!

The reason I didn't make the vertical scrollbar wide enough to click was because the dragger for the SplitPane blocks a lot of it anyway. I never click the scrollbar personally, I only ever really use it as an indicator I can scroll. I'm sure more people use it to scroll than I think so I shouldn't break that. I just noticed `overflow: overlay` works in Chrome so I'm thinking that would be a much better approach since you can click to scroll if you want, it's not always there taking valuable horizontal space and it doesn't interfere with the SplitPane dragger.

As for the the horizontal scroll bars I just thought it looked bad. The `overflow: overlay` takes care of this as well.

I just pushed a commit with the adjustment. Let me know what you think.

crswll:
I can make some gif updates for the README.md (if you like) as well.

yyx990803:
@crswll appreciated - but ideally in another PR? I'll try to take some time to review this this weekend.

crswll:
Absolutely. I'll wait and see how you like the adjustments before making it.

yyx990803:
Tried it out locally and this looks fantastic. Thanks for the great work!

crswll:
Awesome, thanks! Happy to help! I'll get you some updated gifs for the README tonight.

PR #194, opened 2016-12-05
Added eslint-plugin-html as it's required for the build

TheDutchCoder:
eslint-plugin-html is required for a successful build, but was not included in the package.json.


yyx990803:
Thanks, this doesn't seem to be the case if using npm3/yarn with flat npm installs.

TheDutchCoder:
Strange, I'm using npm v4.0.3 (haven't tried yarn yet though)

PR #216, opened 2017-01-04
Fix couple of issues on the events tab after refactor for named slots and namespaced vuex modules

markussorg:
Fixes the following issues on the events tab:

- After the refactor for namespaced vuex modules by @yyx990803 , there was a wrong access of the store in src/devtools/index.js (store.state.app.tab instead of store.state.tab).

- After the refactor for named slots by @yyx990803 , the div for the event data had a missing slot attribute (therefore the event data was not shown).

- Minor styling clean up of the event count bubble.

### DIFF
--- a/src/devtools/App.vue
+++ b/src/devtools/App.vue
@@ -151,7 +151,7 @@ export default {
   overflow hidden
   flex 1
 
-$event-count-bubble-size = 22px
+$event-count-bubble-size = 17px
 
 .event-count
   background-color $active-color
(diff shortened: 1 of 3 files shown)


yyx990803:
Thanks! I'm actually in the progress of making more refactors and have already noticed these in my local branch - merging this conflicts with a few changes I currently, but appreciate the efforts!

P.S. trying to get the events tab shipped in next release.

PR #222, opened 2017-01-06
property "constructor" in data with error fixed.

code1492:
Fixed an error when declaring property "constructor" in data.
image

### DIFF
--- a/shells/chrome/src/detector.js
+++ b/shells/chrome/src/detector.js
@@ -9,7 +9,7 @@ function detect (win) {
     const all = document.querySelectorAll('*')
     const el = [].find.call(all, e => e.__vue__)
     if (el) {
-      let Vue = el.__vue__.constructor
+      let Vue = el.__vue__.prototype.constructor
       while (Vue.super) {
         Vue = Vue.super
       }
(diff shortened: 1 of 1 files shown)


shells/chrome/src/detector.js
```diff
@@ -9,7 +9,7 @@ function detect (win) {
     const all = document.querySelectorAll('*')
     const el = [].find.call(all, e => e.__vue__)
     if (el) {
-      let Vue = el.__vue__.constructor
+      let Vue = el.__vue__.prototype.constructor
```
code1492: It is my mistake. Change 'prototype' to '\_\_proto\_\_'

shells/chrome/src/detector.js
```diff
@@ -9,7 +9,7 @@ function detect (win) {
     const all = document.querySelectorAll('*')
     const el = [].find.call(all, e => e.__vue__)
     if (el) {
-      let Vue = el.__vue__.constructor
+      let Vue = el.__vue__.__proto__.constructor
```
yyx990803: Let's use `Object.getPrototypeOf` instead

yyx990803:
Thanks!

PR #231, opened 2017-01-10
make z-index bigger

andreiglingeanu:
I'm often having components rendered inside nested modals which have big `z-index`-es. It is very inconvenient for me to not be able to quickly hover over an component inside vue-devtools and not get a nice-looking hightlight (it is just hidden behind my modals). This pull request fixes that, probably in a dumb way.


andreiglingeanu:
Thanks!

PR #232, opened 2017-01-10
Fix recognition of Vue root elements if they are in the Shadow DOM

danieldiekmeier:
At the moment, the Vue DevTools will not recognize any Vue root elements if they are in a Shadow DOM.

This is what my DOM tree looks like:
image

The DevTools don't show anything
image

This is because the [`walk` function](https://github.com/vuejs/vue-devtools/blob/master/src/backend/index.js#L124-L133) only traverses `.childNodes`, but not `.shadowRoot.childNodes`.

I added the necessary code to traverse the `.shadowRoot`, and it works fine for me, now.

I also changed the Loop from `Array.prototype.forEach.call` to `node.childNodes.forEach`, because [`NodeList` supports `.forEach()`](https://developer.mozilla.org/de/docs/Web/API/NodeList), and it also works fine.

I didn't find a good/reasonable way to concat the two NodeLists before iterating (`.concat()` isn't available on `NodeList`s), so I duplicated the code, for now.

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -123,7 +123,18 @@ function scan () {
 
 function walk (node, fn) {
   if (node.childNodes) {
-    Array.prototype.forEach.call(node.childNodes, function (node) {
+    node.childNodes.forEach(function (node) {
+      const stop = fn(node)
+      if (!stop) {
+        walk(node, fn)
+      }
+    })
+  }
+
+  // This code is duplicated because there doesn't seem to be
+  // a reasonable way to concat NodeLists
+  if (node.shadowRoot && node.shadowRoot.childNodes) {
+    node.shadowRoot.childNodes.forEach(function (node) {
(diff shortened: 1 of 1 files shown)


src/backend/index.js
```diff
@@ -123,7 +123,18 @@ function scan () {
 
 function walk (node, fn) {
   if (node.childNodes) {
-    Array.prototype.forEach.call(node.childNodes, function (node) {
+    node.childNodes.forEach(function (node) {
```
yyx990803: we can extract the iterator function and reuse it for both lists.

jpommerening: Hi Evan, happy valentines 😂

Chrome just updated me to 3.0.7.
Console now says:
```
Uncaught TypeError: e.childNodes.forEach is not a function
```

I guess this might be the source? Incredible turnaround time though :ship:

Issue #171, opened 2016-11-10
JS error in Chrome dev tools with Prototype js


davemerrill:
With the latest version of Prototype.js loading (1.7.3), vue-devtools enabled, and Chrome dev tools open, I see this error:
    Uncaught TypeError: this.each is not a function(…)
    detect	@	prototype.js:905
    (anonymous function)	@	VM175:1

The 'VM...' reference changes, but is always there. Happens in any file, no js except a link to Prototype. Removing Prototype or disabling vue-devtools avoids the error.

Just to be clear, I'm not using Prototype in Vue projects, but I work on both, and it's slightly unfortunate to have to disable the extension for Prototype work. Not critical, but if the problem is easily avoided, it should be.

Example file attached, nothing in it but a link to Prototype on CDN.
[vue-dev-tools-test.zip](https://github.com/vuejs/vue-devtools/files/583361/vue-dev-tools-test.zip)

euperia:
I'm also having this issue.  Shows on any websites that use prototype.js.  I'm using Vuejs on a Magento site and seeing this on every page load regardless of Vue being loaded or not.

nickgr6:
> Just to be clear, I'm not using Prototype in Vue projects, but I work on both, and it's slightly unfortunate to have to disable the extension for Prototype work. Not critical, but if the problem is easily avoided, it should be.

Same here. Would be nice, if it'll be fixed :)

shershen08:
It could be fixed with following patching the nodes list in the detector.js

```
  let all = document.querySelectorAll('*')
        let prototypeJSSafeNodesList = [].map.apply(all, (el) => {
            el.each = function() {};
            return el;
        })
        const el = [].find.call(prototypeJSSafeNodesList, e => e.__vue__)
        if (el) {
//
```

posva:
So it's prototype.js just breaking because it's overriding prototypes, which, is a known bad practice 😩 ?
@shershen08 I think it'll be better to do a simple for loop instead

shershen08:
@posva that's what prototype.js is about, patching native objects. It's so old that good / bad practices didn't exist at those days I guess. 
Thing with for-loop will iterate all nodes in the dom tree, while find should return first isn't it?

posva:
@shershen08 yes, that's it. You can even create a `find` function. It stops when it finds one occurrence that meets the condition

shershen08:
> You can even create a find function.
[...]

PR #240, opened 2017-01-23
Update detector.js

shershen08:
added findVueRootNode method to avoid prototype.js error
#171

### DIFF
--- a/shells/chrome/src/detector.js
+++ b/shells/chrome/src/detector.js
@@ -5,9 +5,15 @@ window.addEventListener('message', e => {
 })
 
 function detect (win) {
+  function findVueRootNode (array, callback) {
+    for (let i = 0; i < array.length;i++) {
+      console.log(array[i])
+      if (Function.prototype.call(array[i], callback)) return array[i]
+    }
+  }
   setTimeout(() => {
     const all = document.querySelectorAll('*')
-    const el = [].find.call(all, e => e.__vue__)
+    const el = findVueRootNode(all, e => e.__vue__)
     if (el) {
       let Vue = el.__vue__.constructor
       while (Vue.super) {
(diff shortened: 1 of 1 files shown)


shells/chrome/src/detector.js
```diff
@@ -5,9 +5,15 @@ window.addEventListener('message', e => {
 })
 
 function detect (win) {
+  function findVueRootNode (array, callback) {
+    for (let i = 0; i < array.length;i++) {
+      console.log(array[i])
```
bartlomieju: I guess it should not be here

shershen08: oh, totally, it's from my local test setup O_o

shershen08:
is there any more action needed for this change?

posva:
Sorry, I forgot about this. I'll test it and merge this afternoon

shells/chrome/src/detector.js
```diff
@@ -5,9 +5,14 @@ window.addEventListener('message', e => {
 })
 
 function detect (win) {
+  function findVueRootNode (array, callback) {
+    for (let i = 0; i < array.length;i++) {
+      if (Function.prototype.call(array[i], callback)) return array[i]
```
yyx990803: Why `Function.prototype.call` here?

shells/chrome/src/detector.js
```diff
@@ -5,9 +5,14 @@ window.addEventListener('message', e => {
 })
 
 function detect (win) {
+  function findVueRootNode (array, callback) {
+    for (let i = 0; i < array.length;i++) {
```
yyx990803: space after second semicolon please :P

shershen08:
fixed requested changes

shells/chrome/src/detector.js
```diff
@@ -5,9 +5,14 @@ window.addEventListener('message', e => {
 })
 
 function detect (win) {
+  function findVueRootNode (array, callback) {
+    for (let i = 0; i < array.length; i++) {
+      if (callback.call(array[i])) return array[i]
```
yyx990803: No need for the `.call` here either, just `callback(array[i])` would do.

BTW The first argument to `fn.call()` is the `this` context it will be called with.

Issue #84, opened 2016-03-09
dark theme


marcosrjjunior:
Maybe we could create a dark theme

![vuedtools1](https://cloud.githubusercontent.com/assets/5287262/13644769/47dc52ce-e606-11e5-8e1a-1c4abd174381.png)
![vuedtools2](https://cloud.githubusercontent.com/assets/5287262/13644768/47d68aec-e606-11e5-88a7-28e479974e90.png)
![vuedtools3](https://cloud.githubusercontent.com/assets/5287262/13645431/c36b60f8-e609-11e5-9dd0-f43904313ce5.png)

blake-newman:
Feel free to provide a PR, for this. Currently this is not on the roadmap.

Issue #79, opened 2016-03-06
Chrome dark theme


KevinOngko:
![1457243950457](https://cloud.githubusercontent.com/assets/15880638/13552568/75f2d02e-e39b-11e5-8e50-1a908951b0c6.jpg)
it would be great to support the new dark theme :)

PR #243, opened 2017-02-03
Applies a dark theme, based on a global class

jhoff:
~~I need a little help here, but I did most of the the leg work from a styling perspective.~~

~~According to https://bugs.chromium.org/p/chromium/issues/detail?id=608869, there should now be possible to get the value of the chrome devtool theme via `chrome.devtools.panels.themeName`.~~

~~I'm not familiar enough with chrome devtool development to know how to do it correctly, so hopefully this is an easy one for someone that knows what they're doing.~~

~~Implementation of the dark theme is done by adding a `dark` class to the root `#app` element. See change to line 4 at `src/devtools/App.vue`.~~

Edit: When installed in chrome devtools, now detects the theme automatically. Note: the dark theme doesn't work when running `npm run dev`. You will have to add the `dark` class manually to test that I guess.

screenshot from 2017-02-03 00-09-17
screenshot from 2017-02-03 00-09-50
screenshot from 2017-02-03 00-10-13
screenshot from 2017-02-03 00-16-39

Also, any feedback would be appreciated.

### DIFF
--- a/src/devtools/App.vue
+++ b/src/devtools/App.vue
@@ -61,7 +61,7 @@ export default {
     message: state => state.message,
     tab: state => state.tab,
     newEventCount: state => state.events.newEventCount,
-    isDark: state => chrome.devtools && chrome.devtools.panels.themeName === 'dark'
+    isDark: () => chrome.devtools && chrome.devtools.panels.themeName === 'dark'
   }),
   methods: {
     switchTab (tab) {
--- a/src/devtools/App.vue
+++ b/src/devtools/App.vue
@@ -1,7 +1,7 @@
 <style lang="stylus" src="./global.styl"></style>
 
 <template>
-<div id="app" class="app">
+<div id="app" :class="{ app: true, dark: true }">
   <div class="header">
     <img class="logo" src="./assets/logo.png" alt="Vue">
     <span class="message-container">
@@ -102,11 +102,13 @@ export default {
   width 100%
   height 100%
   user-select none
-  background-color #fff
+  background-color $background-color
   display flex
   flex-direction column
   h1
     color #42b983
+  &.dark
+    background-color $dark-background-color
 
 .header
   display flex
@@ -115,6 +117,8 @@ export default {
   box-shadow 0 0 8px rgba(0, 0, 0, 0.15)
   font-size 14px
   position relative
+  .app.dark &
+    border-bottom 1px solid $dark-border-color
 
 .logo
   width 30px
@@ -137,9 +141,11 @@ export default {
   cursor pointer
   position relative
   border-bottom-color transparent
-  background-color #fff
+  background-color $background-color
   color #888
   transition color .35s ease
+  .app.dark &
+    background-color $dark-background-color
 
   &:hover
     color #555
@@ -186,6 +192,8 @@ $event-count-bubble-size = 18px
   position absolute
   right 0
   top 12px
+  .app.dark &
+    background-color $dark-background-color
 
 .active-bar
   position absolute
--- a/src/devtools/App.vue
+++ b/src/devtools/App.vue
@@ -1,7 +1,7 @@
 <style lang="stylus" src="./global.styl"></style>
 
 <template>
-<div id="app" :class="{ app: true, dark: true }">
+<div id="app" :class="{ app: true, dark: isDark }">
   <div class="header">
     <img class="logo" src="./assets/logo.png" alt="Vue">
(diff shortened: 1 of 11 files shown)


jhoff:
This is an implementation of #79 and #84

posva:
Cool! Just make sure it's synced with the devtools colour

jhoff:
Yup I used all of the same background and border colors that the dark theme already uses.

posva:
I meant that the theme should be dark only when the devtools are in dark mode too. But you're right to do so, thanks for pointing that out 👍

jhoff:
@posva Right, which is why this is marked as a [WIP]. I mentioned in the description that I needed some help finishing that part up. I don't know anything really about extension development. It seems that we can determine which theme devtools is currently using through the `chrome.devtools.panels.themeName` property, but `chrome.devtools` is undefined within the scope of `App.vue` when running `npm run dev` ( which makes sense ).

I'll dig into it a little more, but in the meantime if anyone has a recommendation on the best way to implement the switch from light to dark, let me know. I just don't have a lot of bandwidth at the moment.

jhoff:
Ok It wasn't as complicated as I made it out to be in my head. This should be ready to go.

image
image

posva:
Thanks a lot for looking into this.
I need to test this locally and will come back to you asap

crswll:
Sweet. I was just about to start working on this. Looks good.

posva:
Sorry for the delay, testing now. I'll let you know 🙂

posva:
ping @yyx990803

crswll:
My only objection, not that it matters, is this is very hard to read. I'd personally find it frustrating if I used the dark theme regularly (only reason I don't is because of the vue-devtools).

screen shot 2017-02-11 at 1 09 39 pm

yyx990803:
Nice work - as @crswll mentioned, I think we should give the text in the state inspection panel more contrast (especially the purple and the gray base text).

posva:
Maybe using a cyan instead of the purple and orange instead of red or something similar to
screen shot 2017-02-11 at 19 25 03

yyx990803:
Yeah, emulating the color scheme of the native devtool would be great.

jhoff:
I'll get this updated in the next day or so as soon as I get back from vacation

jhoff:
image

How's this look?

posva:
Looks good!
Where, in the elements tab did you find those colors?

jhoff:
Mostly from the properties viewer for HTML elements:

image

andreiglingeanu:
Very nice looking!

src/devtools/App.vue
```diff
@@ -60,7 +60,8 @@ export default {
   computed: mapState({
     message: state => state.message,
     tab: state => state.tab,
-    newEventCount: state => state.events.newEventCount
+    newEventCount: state => state.events.newEventCount,
+    isDark: () => chrome.devtools && chrome.devtools.panels.themeName === 'dark'
```
yyx990803: A few small things:

- The devtools code itself doesn't assume it's always running in Chrome. We need to check `typeof chrome !== 'undefined'` first.

- `isDark` doesn't really need to be a computed property. It's initialized only once and shouldn't be in `mapState` either. Let's move it to `data()` instead.

jhoff: Hows that look?

yyx990803:
Thank you @jhoff , great work! 👍

ackzell:
Hello, 

I just found out a way to integrate vue-devtools into an [NW.js](https://nwjs.io/) dev project. This makes it awesome to have a Vue dev env targeted for nw!

What I found though, is that the dark theme is not being applied to the NW dev tools once I change the theme. I noticed that it has a different menu for changing themes on the settings pane.

NW (0.24.4):
image

Chrome (60.0.3112.113 (Official Build) (64-bit) ): 
image

I think something on this check isn't being matched in the case of the `chrome` object in NW:
https://github.com/vuejs/vue-devtools/blob/master/src/devtools/App.vue#L57

I forced it to be `true` and rebuilt the extension for my use case, that made it appear dark.  

I acknowledge I don't know enough about NW.js yet nor much about how closely it is a "Chrome" window. Much less about creating extensions! Also know that it is a bit of a stretch to ask for a fix on such a thing, I just thought it would be worth sharing my findings.

Cheers!

posva:
@ackzell If anyone finds out how to fix it, we'll welcome a PR 👍

PR #254, opened 2017-02-15
Fix #253

### LINKED ISSUE #253: Uncaught TypeError: e.childNodes.forEach is not a function     backend.js:1
Hi

I am using VueJS and Laravel 5.3, and have been creating some Notifications using Pusher. 
Everything seems to be working ok, except when I use the Chrome devtools and click on the Console tab, I get the following red text:

Uncaught TypeError: e.childNodes.forEach is not a function

So far it hasn't affected the working of any of my functions, but I can not see my components in the devtool Vue tab

I have no childNodes in my code, and I do not have a file called backend.js that I can find, so I think this file is in the devtool.

posva:
use Array.prototype.forEach.call in dom nodes

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -132,13 +132,13 @@ function walk (node, fn) {
   }
 
   if (node.childNodes) {
-    node.childNodes.forEach(iterator)
+    Array.prototype.forEach.call(node.childNodes, iterator)
   }
 
   // This code is duplicated because there doesn't seem to be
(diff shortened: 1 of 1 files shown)


posva:
ping @yyx990803

yyx990803:
Oops, just pushed 4ad6d6b... I guess it's best to just stick to for loops -.- Thanks!

posva:
No worries 🙂

PR #255, opened 2017-02-15
List getters in Vuex tab

AnthonySendra:
Hello,

This PR allows to list all getters defined in the application, just next to the state.

gif0k0igsv

I have a huge application with complex getters, because my state needs to be manipulate (flatten, transformed, ...). Sometimes, it's very difficult to guess the result of the getter according to the state.
In the dev exemple I added a simple getter `isPositive` that returns `true` if `count` is upper or equal than zero.

**Implementation**:
* The action `vuex/importState` doesn't exist anymore. When the user import a state, I send an event through the bridge in order to get updated getters. The bridge will send back the `vuex:init` event to initialize the state and getters ([here](https://github.com/vuejs/vue-devtools/compare/master...AnthonySendra:master#diff-465ca035d2cf1cbe2006f96f847ee38fR33)).
* ^ This also fix a bug: the UI was not updated when the state was imported.
* In the internal store, `base` and `initial ` have now `state` and `getters` as attributes
* On each mutations, the new value of getters is send ([here](https://github.com/vuejs/vue-devtools/compare/master...AnthonySendra:master#diff-465ca035d2cf1cbe2006f96f847ee38fR15))

### DIFF
--- a/test/specs/test.js
+++ b/test/specs/test.js
@@ -150,6 +150,16 @@ module.exports = {
         .assert.containsText('#counter p', '2')
         .frame(null)
 
+      // getters
+      .assert.containsText('.vuex-state-inspector', 'isPositive:true')
+      .frame('target')
+        .click('.decrement')
+        .click('.decrement')
+        .click('.decrement')
+        .frame(null)
+      .click('.history .entry:nth-child(4)')
+      .assert.containsText('.vuex-state-inspector', 'isPositive:false')
+
       // toggle recording
       .click('.toggle-recording')
       .assert.containsText('.toggle-recording', 'Paused')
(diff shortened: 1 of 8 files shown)


test/specs/test.js
```diff
@@ -158,7 +168,7 @@ module.exports = {
       .frame('target')
         .click('.increment')
         .frame(null)
-      .assert.count('.history .entry', 1)
+      .assert.count('.history .entry', 4)
```
AnthonySendra: Because 3 more mutations added

yyx990803:
@AnthonySendra thanks for this PR - looking great at the moment, I just need to pull it locally to try it out. Will probably have to do that after shipping Vue core 2.2 though.

amirrustam:
#262 can be closed if this is merged.

Issue #252, opened 2017-02-15
can vue-devtools support data variable search feature(similar to search component)?


cnweibo:
For a bigger project, there will be pretty much data items, it is a little boring to find a variable and see how it changes during app testing.
Can vue-devtool support a feature of searching data variables similar to searching component?

thanks!

bartlomieju:
I'll happily do it once #261 is merged

posva:
This was actually added in #261

cnweibo:
great, it works as a dream now. Development with vue-devtool and vue is an interesting job!
thanks to vue team

PR #261, opened 2017-02-17
Break props, data and computed in different sections and make them searchable 

### LINKED ISSUE #258: State Inspector: break props/data/computed into separate sections
Open to PRs!

markussorg:
Fixes #258. Needs: npm install (added lodash.groupBy as dependency)

devtools

### DIFF
--- a/src/devtools/views/components/ComponentInspector.vue
+++ b/src/devtools/views/components/ComponentInspector.vue
@@ -18,8 +18,8 @@
       <div class="notice" v-if="target.state && !target.state.length">
         <div>This instance has no reactive state.</div>
       </div>
-      <div v-else>
-        <div class="data-container">
+      <div v-else class="data-wrapper">
+        <div class="data-el">
           <div class="data-type">data</div>
           <div class="data-fields">
             <div v-if="groupedState.undefined">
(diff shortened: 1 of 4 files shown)


andreiglingeanu:
Nice looking!

yyx990803:
Nice work! How does it look when it's narrow? Does it go vertical? :)

src/devtools/views/components/ComponentInspector.vue
```diff
@@ -84,4 +110,20 @@ export default {
...
   margin 0 10px
+
+.data
+  padding: 20px 0px
+
+.data-container
+  float: left
+  min-width: 33%
+  padding: 0px 10px
+
```
crswll: I'd move the `float` and `min-width` for `.data-container` into a media query here so it stacks when we don't have the horizontal space. I'd *try* this. I didn't test it at all but should get you close.

```stylus
@media (min-width: $wide)
  float left
  width 33.333%
  overflow overlay
  ```

markussorg:
Added responsive styling. @crswll: Took a slightly different apprach with flexbox. 

responsive

markussorg:
Added data search. Fixes #252

data_search

PR #264, opened 2017-02-19
Fix components names on Windows

### LINKED ISSUE #256: unnamed component's tag name shows as its file path
## env
- chrome  56.0.2924.87
- vue 2.1.10
- vue-devtools 3.0.8

## screenshot

image

Naming `<ProgressBar>` 
```javascript
{
name: 'progress-bar'
}
```
image

fix this tag.

posva:
Fix #256 

I ended up doing the manual (ugly) fix because I couldn't find a basename/path package that worked on windows...


yyx990803:
Can we remove the diff for yarn.lock?

posva:
@yyx990803 Sure, done. I let it because it wasn't working on windows without that

PR #270, opened 2017-02-28
Clean up styling of dev tools test page and work on folder structure

markussorg:
Adds styling to the test page and reworks folder structure in shells/dev:

vue-devtools-styling


yyx990803:
This is nice but... do the tests still pass? It seems the e2e tests would have to be completely rewritten if this were merged.

michalsnik:
I think it's long overdue and we should probably close it @Akryum 
@armano2 picked this up, so it would be good to continue work in one place.

Akryum:
I think we will need to do a pass after #486

Issue #199, opened 2016-12-12
Inherited properties are not displayed in data inspector


rpkilby:
When using a plain object, both `name` and `_name` are displayed in the data inspector

```js
new Vue({
    el: '#app',
    data: {
        customer: {
            _name: 'John Doe',
            get name() {
                return this._name;
            },

            set name(value) {
                console.log(`name set to: ${value}`);
                this._name = value;
            },
        },
    },
});
```

In this case, the inherited `name` property is missing from the data inspector

```js
class Person {
    constructor() {
        this._name = 'John Doe';
    }

    get name() {
        return this._name;
    }

    set name(value) {
        console.log(`name set to: ${value}`);
        this._name = value;
    }
}

new Vue({
    el: '#app',
    data: {
        customer: new Person(),
    },
});
```

From an implementation standpoint, this might be related to vuejs/vue/issues/3610

ericnilo:
Also whenever I commit/time travel the specific mutation in vuex (devtools) it becomes a plain Object instead of the original instance of the state/data

e.g.

![image](https://user-images.githubusercontent.com/8043652/40339954-36c4f3c4-5daf-11e8-846b-f738de03429d.png)

`state.form` is an instance of Form class and `state.form.errors` is an instance of Error class, after commit/time travel, `state.form` and `state.form.errors` just became an Object instance :( and I get this error:

![image](https://user-images.githubusercontent.com/8043652/40340077-d6ffc2d8-5daf-11e8-9bbe-0e5318a6c0ba.png)

ryuuji3:
I also have this behaviour when rewinding state. My state happens to be a class, but becomes a plain object; which causes a runtime exception when it attempts to call a prototype method which doesn't exist.

jods4:
Sadly this is still the same in devtools 6.0.0-beta.15.

I suppose only owned keys are displayed in the data panel. When using classes, getter/setter properties are found on the object prototype.

This could be mitigated a new menu entry in the 3-dots menu was added to create a global variable, similar to $0 in the DOM explorer, that you could then dump on the console.

PR #274, opened 2017-03-04
change enumeration method of component's computed props

### LINKED ISSUE #271: Bug: Won't display computed from mixins
Hi,
Never happened before, but I noticed that some computed properties are not displayed when they are coming from mixins && no computed object is register in component definition (see screenshots)

#Case 1
screen shot 2017-03-02 at 13 16 03
screen shot 2017-03-02 at 13 11 19

#Case 2
screen shot 2017-03-02 at 13 15 49
screen shot 2017-03-02 at 13 14 14

bartlomieju:
Fixes #271. `Object.keys` does not include properties from prototype chain and when `computed` is not defined on the component, all of the computed properties come from the prototype.

### DIFF
--- a/shells/dev/target/Other.vue
+++ b/shells/dev/target/Other.vue
@@ -6,8 +6,19 @@
 </template>
 
 <script>
+// this computed property should be visible 
+// even if component has no 'computed' defined
+const computedPropMixin = {
+  computed: {
+    computedPropFromMixin() {
+      return null
+    }
+  }
+}
+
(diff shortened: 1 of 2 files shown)


bartlomieju:
I think it might be a good idea to add test assertion and also implement the same mechanism for `data` as there is issue related (#199). I think I'll have some time tomorrow to commit to that.

posva:
@bartlomieju that'd be great

PR #275, opened 2017-03-06
Support Classes as types in props

### LINKED ISSUE #273: Support native classes as types in props
I have a component that contains the following property:

```js
    props: {
      owner: { required: true, type: Owner }
    },
```

In this case Owner is a ES6 class with the same name that I `import` from another file.
Whenever I try to inspect the component i get this error in the console:

```
Uncaught TypeError: Cannot read property '1' of null
    at y (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:4057)
    at chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:3950
    at Array.map (native)
    at d (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:3854)
    at p (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:3334)
    at u (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:2334)
    at t.<anonymous> (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:1524)
    at t.n.emit (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:15051)
    at chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:12718
    at n (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:761)
```

If I remove the type validation, or set it to `Object` it works.

posva:
Fix #273
Also returns any instead of throwing an error if the regex doesn't match
anything.

### DIFF
--- a/test/specs/test.js
+++ b/test/specs/test.js
@@ -24,6 +24,13 @@ module.exports = {
       // should expand root by default
       .assert.count('.instance', baseInstanceCount)
 
+      // prop types
+      .click('.instance .instance:nth-child(2) .self')
+      .assert.containsText('.component-name', 'Target')
+      .assert.containsText('.data-field:nth-child(5)', 'msg:"hi"')
+      .assert.containsText('.data-field:nth-child(6)', 'obj:undefined')
+      .assert.containsText('.data-field:nth-child(2)', 'ins:Object')
+
       // select child instance
       .click('.instance .instance:nth-child(1) .self')
       .assert.containsText('.component-name', 'Counter')
--- a/shells/dev/target/MyClass.js
+++ b/shells/dev/target/MyClass.js
@@ -1,0 +1,5 @@
+export default class MyClass {
+  constructor () {
+    this.msg = 'hi'
+  }
+}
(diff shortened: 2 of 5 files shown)


test/specs/test.js
```diff
@@ -24,6 +24,13 @@ module.exports = {
       // should expand root by default
       .assert.count('.instance', baseInstanceCount)
 
+      // prop types
+      .click('.instance .instance:nth-child(2) .self')
+      .assert.containsText('.component-name', 'Target')
+      .assert.containsText('.data-field:nth-child(5)', 'msg:"hi"')
+      .assert.containsText('.data-field:nth-child(6)', 'obj:undefined')
+      .assert.containsText('.data-field:nth-child(2)', 'ins:Object')
```
posva: I tried to test the type content with `.assert.containsText('.data-field:nth-child(2) .self .prop .meta-field:nth-child(1) .value', 'MyClass')` but the content is always empty although it does appear in the Devtools as `MyClass`

PR #285, opened 2017-03-15
Hide devtools button when not in devtools context

David-Desmaisons:
With this PR, the button "inspect DOM" is not displayed when the vue-devtools executes out of the chrome-devtools context. This improves the user experience in such scenario (such as in [Neutronium](https://github.com/NeutroniumCore/Neutronium)).


michalsnik:
It's not possible to update this PR, so I'll create new with little CSS fix so we can merge it without further delay. Thank you @David-Desmaisons for submitting this PR in the first place :)

Issue #98, opened 2016-04-29
Multiple roots


aemadrid:
It stops working when I have two roots (two separate apps) running on the same page. The left sidebar might show the right information (multiple roots) but clicking on any of the roots or its children (if shown) will not update the data on the right sidebar.

afkatja:
We seem to have this issue, too.

guidobouman:
@yyx990803 Any news on this? Can we support this, or is it just not an option in the current devtools setup?

guidobouman:
Ping @simplesmiler ?

spaceemotion:
Not sure if this is the right issue for this, but I got a minor issue with multiple vue roots on my page.

I built a small modal system that attaches itself as a new vue instance at the end of the body. The devtools only seem to pick it up once I hit its "Refresh" button.

Just something minor, but would be great if there's a way to pick this up automagically.

michalsnik:
I think we can close this issue now @posva 

And if you still experience following problem @spaceemotion it would be great if you create separate issue with an example codebase to reproduce the problem :)

Issue #234, opened 2017-01-11
Vue 1 (1.0.24) multiple Vuejs instances on same page


RedLucas:
For some reason, if I have multiple Vuejs instances on the same page, the Root elements get merged together in the Devtools. (We are rolling Vuejs into our project slowly, so that's why we have multiple instances). 

![screen shot 2017-01-11 at 1 52 06 pm](https://cloud.githubusercontent.com/assets/8796324/21867932/4902c646-d805-11e6-8aa4-9af2044f498f.png)

RedLucas:
Upgraded to 1.0.28 to eliminate the possibility that it's the issue, but nothin...

posva:
If you introducing Vue progressively you should add Vue 2 instead. It works fine with Vue 2.
PRs to fix this are welcome but I think this is probably going to be a wontfix

guidobouman:
@posva We're running Vue2 on our project, and still have this issue. See #98.

guidobouman:
Btw, fixed by https://github.com/vuejs/vue-devtools/pull/292 as well.

michalsnik:
I also think this should rather be a wontfix @posva and we should close this issue, as it works properly with Vue 2 :)

PR #292, opened 2017-03-19
fix instance id clashes with multiple roots on a page

### LINKED ISSUE #289: Clicking the component redirects to different page
Versions:
vue-devtools: 3.1.2
chrome:  56.0.2924.87 (64-bit)
vue: 2.1.10
vuex: 2.1.2

STR:
- Go to a page with 2 different vue instance on it, each instance has at least one component
- Open Vue Devtools and go to the 'Components' tab
- Click on a component in the list

Expected:
- The component (and only that component) will be selected
- The component's scope will appear in the right hand panel

Actual:
- Both the item click on and it's corresponding item in the list for the other Vue instance is selected.
- App redirects to the login page.

This is really weird. Perhaps it's something to do with multiple Vue instances on the same page?

screen shot 2017-03-15 at 2 24 35 pm

bartlomieju:
Closes #289.

Allows for multiple roots using different `Vue` instances (e.g. 1 from webpack bundle, other loaded from CDN) without 'selecting' 2 instances when clicking on an instance in component tree.

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -60,9 +60,7 @@ function connect () {
 
   bridge.on('select-instance', id => {
     currentInspectedId = id
-    console.log(id)
     const instance = instanceMap.get(id)
-    console.log(instance)
     if (instance) {
       scrollIntoView(instance)
       highlight(instance)
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -28,6 +28,7 @@ let bridge
 let filter = ''
 let captureCount = 0
 let isLegacy = false
+let rootUID = 0
 
 export function initBackend (_bridge) {
   bridge = _bridge
@@ -128,7 +129,9 @@ function scan () {
       if (baseVue.config && baseVue.config.devtools) {
         // give a unique id to root instance so we can
         // 'namespace' its children
-        instance.__VUE_DEVTOOLS_ROOT_UID__ = rootInstances.length
+        if (typeof instance.__VUE_DEVTOOLS_ROOT_UID__ === 'undefined') {
+          instance.__VUE_DEVTOOLS_ROOT_UID__ = ++rootUID
+        }
         rootInstances.push(instance)
       }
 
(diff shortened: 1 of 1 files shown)


src/backend/index.js
```diff
@@ -60,7 +60,9 @@ function connect () {
 
   bridge.on('select-instance', id => {
     currentInspectedId = id
+    console.log(id)
```
posva: Hey, you forgot these logs 🙂

bartlomieju: Indeed I did, resolved!

guidobouman:
#98 and #234 are duplicate issues for this problem.

RedLucas:
Thank you for fixing this, you guys do great work.

lucasctd:
@RedLucas Hey, how did you get this to work properly?
Thanks in advance.

lucasctd:
@RedLucas I got it hehe. Thank you guys. :)

guidobouman:
@posva This looks like it's mergeable? The we can close #98 & #234. =]

mika76:
Is this going in any time soon? This bug is really irritating...

posva:
ping @yyx990803

src/backend/index.js
```diff
@@ -126,6 +126,9 @@ function scan () {
         baseVue = baseVue.super
       }
       if (baseVue.config && baseVue.config.devtools) {
+        // give a unique id to root instance so we can
+        // 'namespace' its children
+        instance.__VUE_DEVTOOLS_ROOT_UID__ = rootInstances.length
```
yyx990803: the `scan` function may get called multiple times (when the user force refreshes), in this case `rootInstances.length` may not be stable and existing root instances' ids may get mutated.

I think it's safer to have an incrementing value and only assign root UIDs when the root instance doesn't already have one:

``` js
// top level
let rootUID = 0

// ...
if (instance.__VUE_DEVTOOLS_ROOT_UID__ == null) {
  instance.__VUE_DEVTOOLS_ROOT_UID__ = rootUID++
}
```

bartlomieju: Done!

yyx990803:
Thanks!

PR #293, opened 2017-03-19
display regExp string representation in instance inspector

### LINKED ISSUE #276: RegExp are displayed as empty objects
In the component properties pane, Regex are not displayed:

image

To reproduce: add a data or computed property that holds a Regex.

What's expected: displaying the Regex

image

bartlomieju:
Closes #276.

Unfortunately this is a bit hacky, because stringifying RegExp loses information about actual expression inside, so I called `.toString()` in `sanitize` method.


armano2:
we don't really have to make additional support for this.

https://github.com/vuejs/vue-devtools/pull/319

posva:
I added some commits at #319

PR #300, opened 2017-03-23
fixup 4a50a8af (reference page in vuejs repo rather than neeravp's one)

CircleCode:


posva:
Oh, thanks. I overlooked that

PR #319, opened 2017-05-07
Fix: RegExp are displayed as empty objects.

armano2:
This fix vuejs/vue-devtools#276
`const rawTypeRE = /^\[object (\w+)]$/` is invalid.....
`const rawTypeRE = /^\[object (.+)\]$/` is valid

### DIFF
--- a/shells/dev/target/index.js
+++ b/shells/dev/target/index.js
@@ -27,7 +27,8 @@ new Vue({
   data: {
     obj: {
       items: items,
-      circular
+      circular,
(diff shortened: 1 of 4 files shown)


posva:
They, thanks for the PR! Could you add a test case for this?

posva:
Although I'm not sure if we should go with #319 instead

armano2:
@posva added

posva:
I added some commits at #326

PR #326, opened 2017-05-10
Support regex

### LINKED ISSUE #276: RegExp are displayed as empty objects

posva:
Closes #276

Add support for native types like Regexp. To add a new native type (eg: Symbol), we have to update the replacer function in util

It looks like:
screen shot 2017-05-10 at 15 40 31

screen shot 2017-05-10 at 15 41 35

Which is the same colour of the devtools

### DIFF
--- a/shells/dev/target/Other.vue
+++ b/shells/dev/target/Other.vue
@@ -42,8 +42,7 @@ export default {
           e: undefined,
           f: true,
           g: 12345,
-          h: 'I am a really long string mostly just to see how the horizontal scrolling works.',
-          i: new RegExp('12345', 'ig'),
+          h: 'I am a really long string mostly just to see how the horizontal scrolling works.'
         }
       }
     }
--- a/shells/dev/target/Other.vue
+++ b/shells/dev/target/Other.vue
@@ -43,6 +43,7 @@ export default {
           f: true,
           g: 12345,
           h: 'I am a really long string mostly just to see how the horizontal scrolling works.',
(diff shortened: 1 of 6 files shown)


posva:
The main problem was that the data being serialised, the regexp information is lost upon stringification. Therefore, the regex case is handled in the replacer that keeps the string representation of the regex in a custom format. This very same format can be used with `Symbols` and maybe others

armano2:
I like your code way more 👍 
> To add a new native type (eg: Symbol)

```js
function replacer (key, val) {
  if (val === undefined) {
    return UNDEFINED
  } else if (val === Infinity) {
    return INFINITY
  } else if (val instanceof RegExp) {
    // special handling of native type
    return `[object RegExp ${val.toString()}]`
  } else if (typeof val === 'symbol') {
    return `[object Symbol ${val.toString()}]`
  } else {
    return sanitize(val)
  }
}

posva:
Now that I think of it, it doesn't make sense to have `object` in the string there, it may be better to have something like `native` or `special` or `custom`

armano2:
Right now most of build in types are "printed" as Object() and i think we should support them.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects

posva:
The NaN display is actually a bug. I'll take a look

posva:
Turned out it wasn't being handled

PR #330, opened 2017-05-12
Component inspector header static

anteriovieira:
Keep header static.

Before: 
header-before

After:
header-after

### DIFF
--- a/src/devtools/App.vue
+++ b/src/devtools/App.vue
@@ -160,7 +160,7 @@ export default {
   &.active
     color $active-color
 
-  &.components
+  &:first-of-type
     margin-left auto
 
   .material-icons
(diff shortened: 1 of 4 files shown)


posva:
I think it will be better to have the filter and inspect button right-aligned by letting the component name take the remaining space

posva:
btw, why does the selected tab marker looks off for you?

anteriovieira:
I'll show the image with your suggestions, and then I'll fix the pull request.

> btw, why does the selected tab marker looks off for you?

I've added a margin to improve the preview of the gif, but I'll avoid doing that.

anteriovieira:
Organize the items in the action header

## Updates

- Right-aligned buttons
- Space for title

header-after-organize

- Press ESC to close import
- Right-aligned buttons
- Maybe add the type as title?

press-esc

PR #331, opened 2017-05-12
[Proposal] More Verbose Component Tree

nerdyglasses:
Someone on Stackoverflow was asking if it was possible to show individual component names in the component tree.

Subconsciously I always wanted this as well.
So why not give it a shot.

It works by implementing an optional computed value `instanceName` in the components.

plain_shell_und_vue-devtools_ _working_copy__master_ _0_changed_files_

This could be useful for lists but also for input components to show the name of the input.

```
Like:
<TextInput: first-name>
<TextInput: last-name>
<TextAreaInput: bio>
```

- It has a note in the README.md
- It has tests for the names
- All previous tests still pass

Please let me know if there are any suggestions or critique.

Stackoverflow:
https://stackoverflow.com/questions/40936918/how-to-distinguish-between-component-instances-in-vue-devtools

Best,
Christian

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,23 @@ Currently only a Chrome devtools extension is available.
 5. Check "developer mode"
 6. Click "load unpacked extension", and choose `shells/chrome`.
 
+### Optional Usage
+
+Implement a computed value `instanceName` in your components to make the component tree be more verbose.
+
+The following example could have the component be rendered as
+`<Person: Evan You>` instead of `<Person>`.
+
+```
+export default {
+  name: 'Person',
+  props: ['name']
+  computed: {
+    instanceName () { return this.name }
+  }
+}
+```
+
 ### Hacking
 
 1. Clone this repo
--- a/shells/dev/target/People.vue
+++ b/shells/dev/target/People.vue
@@ -1,0 +1,22 @@
+<template>
+  <div class="people">
+    <h1>People</h1>
+    <p>This demonstrates how having an individual Component Name can be beneficial for lists.</p>
+    <person first-name="Evan" last-name="You" />
+    <person first-name="Gambo" />
+  </div>
+</template>
+
+<script>
+import Person from './Person.vue'
+
+export default {
+  name: 'People',
+  components: { Person },
+  computed: {
+    instanceName () {
+      return 'Open Me'
+    }
+  }
+}
+</script>
--- a/shells/dev/target/Person.vue
+++ b/shells/dev/target/Person.vue
@@ -1,0 +1,23 @@
+<template>
+  <div class="person">
+    Person: {{ fullName }}
+  </div>
+</template>
+
+<script>
+
+export default {
+  name: 'Person',
+  props: ['firstName', 'lastName'],
+  computed: {
+    fullName () {
+      return [this.firstName, this.lastName]
+               .filter(item => item !== undefined)
+               .join(' ')
+    },
+    instanceName () {
+      return this.fullName
+    }
+  }
+}
+</script>
--- a/shells/dev/target/index.js
+++ b/shells/dev/target/index.js
@@ -4,6 +4,7 @@ import Target from './Target.vue'
 import Other from './Other.vue'
 import Counter from './Counter.vue'
 import Events from './Events.vue'
+import People from './People.vue'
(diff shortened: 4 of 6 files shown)


posva:
I really don't like the need of creating a useless computed property with a specific name to make it work. However, displaying the `key` attribute may be helpful

nerdyglasses:
Hey @posva,

thx for the comment.

You suggested using the `key` attribute. That's actually a pretty good idea!

I'll try to justify why I would still say something like the PR could be beneficial. Maybe we can change the implementation a bit.

There are 3 problems with this from my understanding.

## Problems with having a `key` only implementation

**There is a visual disconnect between what's shown on the browser vs what the component tree would show.**

Demonstration with `key` only
```
Browser shows:
Choose your favorite Marvel character:
- Hulk
- Spiderman
- ...

Component Tree shows:
- <Hero: 1241>
- <Hero: 5235>
- ...
```

I would like it much better if I was able to have look at this:
```
Browser shows:
Choose your favorite Marvel character:
- Hulk
- Spiderman
- ...

Component Tree shows:
- <Hero: Hulk>
- <Hero: Spiderman>
- ...
```

**We wouldn't be able to use this outside from loops.**

It wouldn't be possible to use it for custom input components like this:

```
Form Component
<my-input name="first-name" v-model="firstName" />

Component Tree shows:
- <Hero: Hulk>
- <Hero: Spiderman>
- ...
```

Having the `key` only solution wouldn't justify this feature in the first place in my opinion.

I'm not sure if that that implementation justifies having 
Since vue aims to be open I would say this is against the believes. 

2) `key` will most likely contain an id. I would find it better if I was able to set it to something that I can see in the browser so I can relate to the rendered result better.

There is a visual disconnect with this:



maybe we can have either or?
The `key` attribute would only work in loops.




Let me explain my point of view:
- If it is set then it isn't useless. Its purpose is to make the component tree readable.
- The computed property is optional, there is no need to use it.

posva:
the real problem with key is actually the same that the computed property: we cannot force the developer to add a property to the component that only has an impact in devtools. If people start using the key attribute to simply have a more precise display on the devtools, that's misusing the key attribute.

> I think I might get why you would consider a custom computed value "useless". It mixes component business logic with vue tooling logic. If that's the reason then I totally get it and I'm against it too.

About the other changes, there's a similar problem, we're adding code that has no impact in the app, it shall be removed when bundling or something else and it's not worth the cost... Even if it's call `$debugName` and we point out it's helpful when debugging, it shall also be removed from production code.

nerdyglasses:
1) Do you understand the need for this?
2) Do you have any suggestions how to get there or close to that?

posva:
yes, distinguish same components in dev tools to make it easier to select them
yes, the key (a name can be an id too), so it can help in some cases. Hovering will highlight the actual element in the dom too

nerdyglasses:
> Hovering will highlight the actual element in the dom too

that's true. okay I guess we could wait for other opinions or ideas how to make it work.

I'm not super familiar with the vue core so I might not come up with the best possible solution at this point.

Thx for your opinion again

nerdyglasses:
Hey @armano2 thx for the thumbs down. Can you explain why you don't like this?

Best,
Christian

nerdyglasses:
Here is an example how I use it atm.

screenshot 2017-05-22 15 42 27

armano2:
Hello, right now we already have name property for component's I will prefer to see that name is reactive with possibility to pass function and let ppl do whatever they want there instead forcing us to make computed property with no side effect.

nerdyglasses:
The computed property has been discussed already. We all agree it shouldn't go into the computed properties because that's where the business logic is and we don't want to mix it up with the tooling.

The latest suggestion is that it could be next to the `name` option.
I would argue that it is related to the component like the `name` option is as well.

```
export default {
  name: 'Hero',
  instanceName () {
    return this.name
  },
  computed: {
    // regular computed values.
  }
}
```

nerdyglasses:
But it's becoming more obvious to me that nobody really cares about this. So maybe we close that and call it a day.

It was still interesting to me to make it work and I'll still use it in our project. Unfortunately I'm not sure how to make the version above work. So I'm going to keep the computed properties.

PR #332, opened 2017-05-19
Rotate refresh icon

anteriovieira:
Maybe it is not the best way to implement, but it would be nice to have a behavior that indicase the refresh.

refresh-icon

### DIFF
--- a/src/devtools/App.vue
+++ b/src/devtools/App.vue
@@ -34,7 +34,7 @@
     <a class="button refresh"
       @click="refresh"
       title="Force Refresh">
-      <i class="material-icons">cached</i>
+      <i class="material-icons" ref="refresh">refresh</i>
       <span class="pane-name">Refresh</span>
     </a>
     <span class="active-bar"></span>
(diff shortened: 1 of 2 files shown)


src/devtools/App.vue
```diff
@@ -78,7 +79,13 @@ export default {
       }
     },
     refresh () {
+      this.refreshRotate += 180
+      const refleshIcon = this.$el.querySelector('.refresh-icon')
```
posva: you can use a ref on the element instead

src/devtools/App.vue
```diff
@@ -78,7 +79,13 @@ export default {
       }
     },
     refresh () {
+      let refreshIcon = this.$refs.refresh
+      this.refreshRotate += 180
+
       bridge.send('refresh')
+      bridge.on('flush', () => {
```
posva: I forgot to say that the bridge.on should be setup only once

anteriovieira:
Hi @posva , what do you think about changing the icon? An idea would be to leave the same chrome refresh icon.

<p align="center">

</p>

src/devtools/global.styl
```diff
@@ -95,3 +95,17 @@ $arrow-color = #444
     text-align center
     padding 0.5em
     margin 0 auto
+
+@-moz-keyframes rotate
+  0%
+    -moz-transform rotate(0deg)
```
posva: No need for prefixes because stylus will add them. So keeping the `@keyframes` rule with only `transform` is enough

anteriovieira:
@posva , what do you think of these latest changes?

Issue #57, opened 2015-11-28
Can you please provide a similar extension for firefox ?


neeravp:
Kudos!! The extension is a must while learning Vue atleast. I would like to request you to provide a similar extension for Firefox.

Please! can i expect it?

### Update: (17-March-2017)  
It is now possible to use (chrome's) vue dev-tools in Firefox.  
  
1. Install Firefox Developer Edition  
2. Install Chrome Store Foxified extension  
3. Now go to Chrome Web store and search for vue- devtools, you will now see an  **Add to Firefox** option   
4. Once you click on `Add to Firefox` you will get a pop up with different options, choose **Save Signed Addon** option and save the file to your disk.  
5. Now navigate to `about:addons` in Firefox and click on the ***gear*** icon in the top right corner (just before the search field), you will see an option **Install Addon from File** in the dropdown menu.  
6. Click `Install Addon from file` and choose the file which you downloaded and saved in **Step 4**.  
7. The `vue-devtools` will get installed in Firefox, restart Firefox and start using **vue-devtools** in Firefox.

yyx990803:
It's definitely planned in the future, however probably not until a few months out...

neeravp:
Thanks. Eagerly awaiting.

skyronic:
I've started work on a firefox version of the same. You can check out [my branch](https://github.com/skyronic/vue-devtools/tree/firefox) hope to finish this in 2-4 weeks. 

Progress so far:

![2015-11-30-173650_1920x1080_scrot](https://cloud.githubusercontent.com/assets/80173/11472287/7ad2dc2a-9791-11e5-93bc-c6d4408ce6bb.png)

yyx990803:
@skyronic very cool! Excited to have someone working on this. Let me know if you have any questions :)

neeravp:
Awesome  @skyronic , I will check out your branch.

nacr-dev:
@yyx990803 any news on this?

PascalAOMS:
Almost half a year later. :) Any news?

neeravp:
@yyx990803 , @skyronic  - any progress? still awaiting devtools for firefox

ptibbetts:
throwing in my vote as a firefox developer edition user

j3j5:
Is there any news on this? I'd love to be able to use FF dev edition for Vue just as I do with everything else. @skyronic? Any plans on finishing this or at least releasing some basic instructions to install what you already had?

Thanks!

[...]

PR #339, opened 2017-05-30
Add the extra slash to the file:///* match pattern

clarkbw:
This is part of the work required to make #57 happen (as a pure WebExtension in Firefox)

The `file://*` match pattern is supposed to have 3 slashes so its actually `file:///*`.  

Chrome ignores the issue but their docs do ask for it:  https://developer.chrome.com/extensions/match_patterns

MDN has a specific warning about this: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Match_patterns#Invalid_match_patterns

Tested `file:///.../todomvc/examples/vue/index.html` in Chrome and Firefox.

### DIFF
--- a/shells/chrome/manifest.json
+++ b/shells/chrome/manifest.json
@@ -32,7 +32,7 @@
   "permissions": [
     "http://*/*",
     "https://*/*",
-    "file://*"
+    "file:///*"
   ],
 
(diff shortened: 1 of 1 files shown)


posva:
Oh, thanks! I tested and it works fine

PR #340, opened 2017-05-30
Updated the Roboto woff2 file with new hinting

clarkbw:
Also part of the work to make #57 happen.

Firefox doesn't seem to like the version of Roboto being shipped.  This updates that to the latest version, pulled from here: https://github.com/FontFaceKit/roboto/blob/gh-pages/fonts/Regular/Roboto-Regular.woff2 And now Firefox seems much happier.

Here's a set of comparison screenshots:

Firefox (Original) | Chrome (Original) | Firefox | Chrome
------------ | ------------- | ------------ | -------------
firefox original | chrome original |  firefox updated |  chrome updated

Also, here's a screenshot of the error as seen in the Firefox Browser Console:

screen shot 2017-05-30 at 5 08 46 pm


yyx990803:
Thanks!

PR #341, opened 2017-06-01
Firefox WebExtension support

clarkbw:
This is the last set of pieces to get #57 closed and is part of the followup work to #340 

This PR needs to wait on Firefox to land the work [bug 1300590](https://bugzilla.mozilla.org/show_bug.cgi?id=1300590) such that `$0` and `inspect` bindings will work in the `devtools.inspectedWindow.eval` API.  However the patch in bugzilla is in the final throws so I wanted to get these changes up and available.

### DIFF
--- a/shells/chrome/devtools-background.html
+++ b/shells/chrome/devtools-background.html
@@ -1,1 +1,2 @@
+<meta charset="utf-8">
 <script src="./build/devtools-background.js"></script>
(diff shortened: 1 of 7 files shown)


posva:
Oh cool, thanks for the PR. I'll have to check it once the patch is released, though. Unless there's a way to check it before that 😄

clarkbw:
> Unless there's a way to check it before that

I can add a link in here once the Nightly builds are available!

clarkbw:
Ok, latest [Firefox Nightly](http://nightly.mozilla.org/) has the APIs landed.  Other than #340 I don't think there's much else to do here.

Issue #324, opened 2017-05-10
Feature request: Collapsed Vuex stores by default option


b44rd:
Hi, and thanks for making a fantastic tool - Vue.js and this devtool sure makes web development fun again. 

One thing I could like to see get implemented, is that when a lot of Vuex store modules is listed, it is kind of time consuming to browse them, since the modules are expanded by default. 

My personal opinion is that it could be better to have it collapsed by default, but I guess not all use multiple Vuex stores. So perhaps an alternative could be to have them expanded if only one store module, and collapsed by default if several?

Issue #123, opened 2016-07-08
Config option to set auto-expand to false


jlmmns:
It would be very helpful if we could disable the auto-expand feature.

A lot of my components inherit a rather large object, which appears at the top.

Because of this, I have to scroll down a lot, or collapse it every time.

neeravp:
I totally agree and wish that such a config option is made available.

zackcote:
Yeah this is huge for me, would be a great help!

Renol7:
I agree with this feature request! When there is a components with some large objects it is **very difficult** to see the data. 
Also, if I **refresh** my browser after collapsing vue-devtools objects, they return to auto-expand state.

posva:
I prefer having them expanded but I think we can leave them collapsed if there's a certain amount of properties being displayed. I think it's better to have a filter input in the vuex state too, though

DannyFeliz:
I really need this feature x.x, every time I'm browsing in my components data I have to collapse the arrays and object, would be very cool if this feature is introduced in the next release

ellisio:
We're building a large scale application with Vuex. At this time we have roughly 24 store objects. When I want to quickly scroll and view the payload, or find an object, there is a lot of slowish scrolling so I don't bypass the object I'm looking for.

I'm also one that constantly collapses my directory tree in my IDE and clicks back down to the file(s) I'm working on. It's something with my OCD.

Having this feature would make my work flow a lot less frustrating :)

posva:
Added at #328

PR #347, opened 2017-06-06
Make Vuex store objects collapsed by default

b44rd:
Fix for https://github.com/vuejs/vue-devtools/issues/324 / https://github.com/vuejs/vue-devtools/issues/123

Note: This makes all values collapsed by default. Changing this line makes it way more trivial to work with a project with a lot of Vuex modules 👍

### DIFF
--- a/src/devtools/components/DataField.vue
+++ b/src/devtools/components/DataField.vue
@@ -64,7 +64,7 @@ export default {
   data () {
     return {
       limit: Array.isArray(this.field.value) ? 10 : Infinity,
-      expanded: this.depth === 0 && this.field.key !== '$route' && (subFieldCount(this.field.value) < 5)
+      expanded: this.depth === 0 && this.field.key !== '$route' && (subFieldCount(this.field.value) < 1)
     }
   },
   computed: {


posva:
the 5 there was intended to collapse when there're more than 5 items. I forgot to close the issue but this was added at #328 
Why do you want to collapse everytime?

b44rd:
The main reason I think collapsed by default is better, is for the sake of consistency, plus it makes a store with x modules more inspectable. 

When collapsed by default, it makes the following headlines very easy to scan;
* state
* getters
* mutation

I can then easily navigate to the part I need to debug. 

However, at the moment, some stores are collapsed, and some is expanded. To me, this looks messy and inconsistent. It also makes it harder to keep track of actual changes while having the inspector open. 

Having some open, and some closed, based on the length of the keys also seem like a temporary hack to me. The reason why I simply changed the number, was by the way so other stuff could be plugged in on top of this change later, by passing this number as a variable. 

An alternative solution, which I by the way think is the number one alternative in all cases, would be an option saying "collapse all", or "expand all". 

What do you think?

b44rd:
Another reason, by the way - When working with multiple stores, you typically only work with one store at the time. Getting stuff that don´t require your immediate attention out of your mind, is often a very nice thing. 👍  I´ve been using my branch instead of master since this commit, and the hiearchy (state - getters - mutation) is working way better this way. Ideally, mutation should have been on top, as some people say, but that leads to a jump when the length of the mutation varies. So in the current design of the debugger, having it on bottom while having store properties collapsed seem like the best option to me.

posva:
I don't think it's a matter of consistency but more a question of usage DX. Can you share some screenshots?
If you have few items in your state that can be displayed, you want them displayed by default. If on the other hand, the state doesn't fit the view, you better collapse it. The 5 there looks like a nice compromise that gives a fast way to check that but we may be wrong if there're many modules.

b44rd:
So, here´s two screenshots of the Vuex tab, showing a sample project with 5 Vuex modules. 

Here´s when I open the Vuex tab in vuejs:master

vuejs:master

While here´s what it looks like after this pull request

b44rd:master

This is a simple project that is barely started, but when working in a project with 10+ Vuex modules, the difference becomes way bigger.

b44rd:
Here´s an example of how it looks like with a project of 15 store modules (that by the way will be refactored to 11). But still, it illustrates the problem way better than the 5 modules example. 

Master

vuejs:master

Pull request branch

b44rd:master

Issue #51, opened 2015-11-17
a version for safari will be available soon ?


stephanedemotte:
Safari is much faster than chrome, it could be great to have this extension :)

Thanks !

amirrustam:
I would also love to just use Safari, because it allows me to inspect apps running within iOS Safari, but from my quick search it doesn't look like there is an API to extend the Safari Web Inspector. Chrome has a very nice API for extending their inspector: https://developer.chrome.com/extensions/devtools

So I think Vue DevTools for Safari would just need to be an independent Safari extension and built like so: https://developer.apple.com/library/safari/documentation/Tools/Conceptual/SafariExtensionGuide/Introduction/Introduction.html

yyx990803:
The devtool's own code is browser-agnostic, it all talks through a standard events interface. However I've only got time to make the Chrome version. Theoretically you can build shells for FF/Safari as long as the extension runs as an HTML5 app. Feel free to contribute if anyone is interested.

amirrustam:
@yyx990803 I'm gonna work on this as soon as I get some extra time to dedicate to it. I've done some Safari extensions in the past. I'll report back here. 

Suggestions and tips are welcome in the mean time.

yyx990803:
I probably need to write some basic docs explaining the design and how things work, but the simplest way to start is looking at `shells/dev` which is basically a minimal shell example using an iframe.

amirrustam:
Some docs will be nice and helpful :+1: 

Once I figure out the Safari extension then I can easily do a Firefox version: https://developer.mozilla.org/en-US/docs/Tools/Adding_a_panel_to_the_toolbox

So then we can all live in peace and harmony on Earth. :)

amirrustam:
Spoke too soon, looks like someone has already got the FF version down #57

mikebronner:
@amirrustam have you made any progress on the Safari plugin? Would love that as well. :) I've run into the issue of Safari not allowing extending of the dev-pane as well. :(

kartsims:
@amirrustam Looking forward to peace and harmony on Earth ! (until then, Safari dev tools would also be nice)

mesqueeb:
I would love a Safari plugin! Because often my website works on Chrome, but not on Safari. It's really hard to find out what's exactly bugging in Safari!!! Any hints?

amirrustam:
[...]

PR #352, opened 2017-06-13
Add safari extension

jaredhobbs:
Addresses #51 

Adds a popover toolbar button that loads the devtools.

screen shot 2017-06-13 at 17 10 32


nickmessing:
Wow, good job

yyx990803:
@jaredhobbs thanks a lot for the PR! I tried it locally (merged against latest `master`) but the popover is always blank, tried in both 10.1.1 and the Technology Preview - any ideas?

jaredhobbs:
I'm on Safari 11 in the High Sierra beta. I developed it by running with `npm run dev:safari` then reloading the extension whenever the js rebuilt. The extension is set to only run on localhost and 127.0.0.1. You can open the web inspector in the popover by right clicking the popover window and clicking inspect element.

yyx990803:
Can confirm it works when built in dev mode, however with `npm run build:safari` it shows up empty. Inspecting the popover doesn't seem to provide any useful info :(

jaredhobbs:
:(

I'll keep on hacking to see what's going on in production mode.

jaredhobbs:
ok, `npm run build:safari` works now

src/bridge.js
```diff
@@ -3,13 +3,14 @@ import { EventEmitter } from 'events'
 export default class Bridge extends EventEmitter {
   constructor (wall) {
     super()
-    this.setMaxListeners(Infinity)
-    this.wall = wall
+    const self = this
```
posva: why did you add this?

jaredhobbs: that's what was causing the error in the safari build: `ReferenceError: Cannot access uninitialized variable.`

Might be related to https://bugs.webkit.org/show_bug.cgi?id=171543

posva: Ok, let's add a comment then 🙂

jaredhobbs: done

yyx990803:
Thanks!

thecrypticace:
Might be worth noting in the docs that you will get a blank popover (with console errors) when not running against 127.0.0.1 or localhost (unless you update the extension permissions in the builder).

Perhaps it would be good to trap the error for `safari.application.activeBrowserWindow.activeTab.page` being undefined and throw up a message?

thecrypticace:
Also, holy mackerel this is awesome!

tremendus:
@jaredhobbs This didn't work for me on Sierra with Safari 10 when running on localhost with `npm run build:safari`. Blank popover as reported by Evan.

Is 11 and/or High Sierra and/or dev mode a requirement or should this have worked on Sierra + 10 without dev?

thecrypticace:
I'm running this on Sierra & Safari 10. Right click the popup and open the console. Do you see any errors?

jaredhobbs:
Also make sure you're running the latest code from this repo and the site running at localhost is in development mode

tremendus:
So interestingly I reloaded the page a bunch of times before I took of for dinner and it wasn't working, but I come back few hours later - no changes to anything - and it works just fine. 

Nevermind, and thanks.

ruchern:
Hi Jared,

Can I just check and confirm that there is no way to add a tab to the Safari developer tools along side "elements, console, etc..."?

jaredhobbs:
Unfortunately, no, at least as far as I could tell

bigsee:
@jaredhobbs firstly, thank you for your awesome work here - so useful! 👍🏽

The extension seems to be working on a production site in Safari, even though I have disabled in my application. Same URL in Chrome gives the usual error about DevTools being disabled in production and does not display the devtools tab in the inspector.

Is this expected behaviour? i.e. will Safari devtools always work even against sites in production?

jaredhobbs:
It shouldn't. If you had opened the tools on a dev site and then navigated to a prod site, the tools will still display info from the dev site. It's a known bug and should clear out but doesn't.

bigsee:
@jaredhobbs thank you for the swift response. :) I can confirm this behaviour after testing so I'll leave open the issue I created here: https://github.com/vuejs/vue-devtools/issues/632

PR #357, opened 2017-06-21
Checks that document.doctype is an HTML doc before injecting

### LINKED ISSUE #356: [Firefox] The xml prettyprint disappears
When the extension is activated in Firefox, the prettyprint of the xml documents is lost, it is no longer possible to fold and unfold the nodes, read attributes...

For a few moments if the native display is visible but a moment later it disappears and only show the textnodes values.

To test you can visit the following examples:
- http://qa-dev.w3.org/wmvs/HEAD/dev/tests/xml-nodoctype-wf.xml
- http://impressive.net/people/gerald/2000/10/18/test-staroffice-doc.xml

clarkbw:
This fixes #356

The document.doctype should always be defined for HTML documents.  See https://www.w3.org/TR/DOM-Level-3-Core/core.html#ID-B63ED1A31

Guarding against this should protect from injecting the content scripts into non-HTML pages.

Tested in Firefox and Chrome.

### DIFF
--- a/shells/chrome/src/detector.js
+++ b/shells/chrome/src/detector.js
@@ -27,8 +27,10 @@ function detect (win) {
   }, 100)
 }
 
-// inject the hook
-const script = document.createElement('script')
-script.textContent = ';(' + detect.toString() + ')(window)'
-document.documentElement.appendChild(script)
-script.parentNode.removeChild(script)
+// inject the hook, avoid non HTML pages
+if (document.doctype !== null) {
+  const script = document.createElement('script')
+  script.textContent = ';(' + detect.toString() + ')(window)'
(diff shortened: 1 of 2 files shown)


clarkbw:
This isn't ready for merge, I don't really like the solution. Let me look into the Firefox issue more before falling back on this.

posva:
ok, thanks a lot for taking a look!

yyx990803:
#385 addresses the same issue and `instanceof` seems to be a stricter check - does that look like a somewhat ok solution?

posva:
ping @clarkbw

clarkbw:
Yes, that looks like a much cleaner solution!  💯

PR #385, opened 2017-08-03
Fix inject problems

### LINKED ISSUE #356: [Firefox] The xml prettyprint disappears

dr-itz:
* Don't leak variable "script"
* Fix XML view in FF. This fixes #356 and is similar to #357 but uses an "instanceof"

### DIFF
--- a/shells/chrome/src/detector.js
+++ b/shells/chrome/src/detector.js
@@ -28,7 +28,9 @@ function detect (win) {
 }
 
 // inject the hook
-const script = document.createElement('script')
-script.textContent = ';(' + detect.toString() + ')(window)'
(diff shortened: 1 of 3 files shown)


yyx990803:
Thanks!

PR #413, opened 2017-09-15
Filter events by instance name too + Events tests

eigan:
- Not case insensitive (components tab is)
- No idea what happens if instanceName is `null` or `undefined`


Akryum:
Tested and working great - I have added some e2e tests for the Events tab.

PR #418, opened 2017-09-25
add filtering for Vuex state

### LINKED ISSUE #401: Filter state properties like filtering components
**Problem**: I have a complex and deep nested Vuex store. I often track those nested properties. The problem is now, every edit to the store requires a full page reload and the opened Vuex state is gone.

**Feature request**: It would be cool to search the store for properties like you can search components to quickly watch the desired nested property.

Might be a duplicate of https://github.com/vuejs/vue-devtools/issues/259, but the referenced issue with the referenced merge request didn't fulfill the feature request.

bartlomieju:
Another simple PR that is really helpful, being able to filter out Vuex state. Closes #401


src/devtools/views/vuex/VuexStateInspector.vue
```diff
@@ -50,12 +54,39 @@ export default {
...
   },
-  computed: mapGetters('vuex', [
-    'inspectedState'
-  ]),
+  computed: {
+    ...mapGetters('vuex', [
+      'inspectedState'
+    ]),
+    filteredState() {
+      const filtered = {};
```
posva: What do you think about reusing what is in ComponentStateInspector? So we just need to build the array of properties before filtering them:

```js
const inspectedState = [].concat(
        ...Object.keys(this.inspectedState).map(
          type => Object.keys(this.inspectedState[type]).map(
            key => ({
              key,
              type,
              value: this.inspectedState[type][key]
            })
          )
        )
      )

      return groupBy(sortByKey(inspectedState.filter(el => {
        return searchDeepInObject({
          [el.key]: el.value
        }, this.filter)
      })), 'type')
```

It would probably be nice to refactor the last part in the future so it's reused

bartlomieju: Good idea, I'll do it first thing when I have some spare time.

posva: Don't worry, I can add it myself as well. Thanks a lot for taking the time to contribute to the devtools! 😄

bartlomieju:
@posva I've actually used your snippet - works perfectly

yyx990803:
@posva have you tried this locally?

posva:
@yyx990803 Not yet, let me give it a shot

posva:
Tested and works nicely, ofc, it would be better with a test 😄

posva:
hhm, got some tests failing. I see we haven't turned on ci on prs but I cannot see the project on circle ci either 😅

Akryum:
Tested and working fine :+1:

posva:
Thanks for picking up @Akryum !

PR #448, opened 2017-11-07
fix(safari): update Info.plist to make it work again

posva:
This was necessary to make it work with safari again. I also added your name @yyx990803 to the plugin
BTW, when running in dev mode, the devtools are unable to focus on the actual devtools (like they do in chrome)

### DIFF
--- a/shells/safari/Vue.js devtools.safariextension/Info.plist
+++ b/shells/safari/Vue.js devtools.safariextension/Info.plist
@@ -3,9 +3,9 @@
 <plist version="1.0">
 <dict>
 	<key>Author</key>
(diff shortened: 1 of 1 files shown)


michalsnik:
Since we're most likely are going to drop support for safari, once remote devtools is ready. Can we close it @posva?

posva:
I think we're better off merging it until remote devtools are ready. @yyx990803 could you give this a quick look?

michalsnik:
@posva remote devtools are already out there :) Can we close it? And I can prepare another PR in which we'll drop safari support. cc @yyx990803

posva:
Sure!

PR #454, opened 2017-11-12
Change layout depending on available space

crswll:
Adding the ability to split the panes horizontally or vertically based on the users preference.

### DIFF
--- a/src/devtools/App.vue
+++ b/src/devtools/App.vue
@@ -36,11 +36,6 @@
       title="Force Refresh">
       <i class="material-icons" ref="refresh">refresh</i>
     </a>
-    <a class="button view icon-only"
-      @click="switchView"
-      title="Switch View">
-      <i class="material-icons view-icon" :class="{ 'rotate-90': view !== 'vertical' }">view_stream</i>
-    </a>
     <span class="active-bar"></span>
   </div>
   <component :is="tab" class="container"></component>
@@ -91,10 +86,10 @@ export default {
         refreshIcon.style.animation = 'rotate 1s'
       })
     },
-    switchView () {
+    switchView (mediaQueryEvent) {
       this.$store.commit(
         'SWITCH_VIEW',
-        this.view === 'vertical' ? 'horizontal' : 'vertical'
+        mediaQueryEvent.matches ? 'vertical' : 'horizontal'
       )
     },
     updateActiveBar () {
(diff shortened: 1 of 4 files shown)


anteriovieira:
Hi @crswll , you can provide a gif to demonstrate the behavior?

crswll:
Sure.

Akryum:
Thanks for your PR @crswll :+1: 
What if we switch the layout automatically depending on the width of the devtools?

crswll:
No problem! I'm happy to help.

I didn't do it with media queries out of personal preference. There's was too many times in devtools I'd want a wide view but then it'll turn into two columns instead of one before i changed the "Panel Layout" setting.

I wonder if I can read that setting and honor it and use media queries if it's set to auto.

Feel all right about that?

Akryum:
Why not, we can try a 3-state button with h/v/auto then.

crswll:
Oh no I meant just getting the setting directly from Chrome devtools so I'd remove the button all together and change a class based on that setting.

Akryum:
Oh okay 👍

crswll:
I gave it a quick look and a long Google but I don't think that setting is exposed like the theme unfortunately.

Any of you know if this setting is exposed?

screen shot 2017-12-24 at 9 51 08 pm

`chrome.devtools.panels.themeName` exists but `chrome.devtools.panels.panelLayout` or the like does not seem to.

I'll go with the 3-state button if there's no luck here.

crswll:
I'm finally getting around to this and not digging a three state button. I'm just going to use a media query and have it work "auto" until I figure out how to read that setting and honor that.

crswll:
Ok, now 2 columns when we have over 685px, which is the breakpoint Chrome devtools seems to use.

src/devtools/App.vue
```diff
@@ -31,11 +31,10 @@
       <span class="pane-name">Events</span>
       <span class="event-count" v-if="newEventCount > 0">{{ newEventCount }}</span>
     </a>
-    <a class="button refresh"
+    <a class="button refresh icon-only"
```
Akryum: Why setting icon only here?

crswll: I removed the button that toggles the layout and that had it as well. I can change this to a specific class and remove this more utilitarian class.

Akryum: I mean, we can make the Refresh button normal (by removing this class) now that you removed the other icon.

crswll: All right. It's been removed along with the extra class.

crswll:
Quick shot of it working.

PR #468, opened 2017-12-10
 inspect component from element panel

zzuligy:
now,vue devtool lack the ability choose element,then locate the correspond component.
so it is.


JS-mark:
test

PR #469, opened 2017-12-12
give the vue dev tool the ablity to select the correspond component w…

zzuligy:
provide vue dev tool the ablity to select the the correspond component when select a page element  by inspector.
when the page has lots of component,this function can let the develop find the correspond compoent more fastly.

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -59,17 +59,46 @@ function connect () {
     }
   })
 
-  bridge.on('select-instance', id => {
+  // when select element in dev tool element panel
+  //find the selected elment and select it in VUE component tree
+  bridge.on('element-inspected', uuid => {
+    function findParentComponent (instanceMap, uuid) {
+      let instanceContainer = []
+      instanceMap.forEach((instance, id) => {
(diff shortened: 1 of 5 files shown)


src/backend/index.js
```diff
@@ -59,17 +59,46 @@ function connect () {
...
+      let instanceContainer = []
+      instanceMap.forEach((instance, id) => {
+        let curSelectedDom = document.querySelector(`[data-v_track_id="${uuid}"]`)
+        let isAncestorInstance = instance.$el.contains(curSelectedDom)
+        if (isAncestorInstance) {
+          instanceContainer.push(instance)
+        }
+      })
+      //sort ancestor
+      instanceContainer.sort(function (m, n) {
```
posva: What is the sort for here?

posva:
do you mean #479 ?

zzuligy:
yes.we do the same thing.but his code is more elegant.for example the contextMenu,the __vue__, the exapnd of the selected component

Akryum:
I guess we will go with #476 and #479.

Issue #239, opened 2017-01-22
Provide a setting to choose component tag name between CamelCase and kebab-case


rainfore:
I give my components tag name in kebab-case like 'v-list-view', but it looks ugly ('VListView') in components panel of devtools.
Is there any way to set component tag name in kebab-case?

departure:
Yes, it would be great.

PR #470, opened 2017-12-12
Add option to show original component names

manico:
As requested in #239


posva:
I think this should be moved in the future in some kind of settings panel but looks good imo:

screenflow

PR #471, opened 2017-12-14
fix event hook

maxushuang:
Once an event, the event will not be removed. Because this event name become `$$event` in the off function.

### DIFF
--- a/src/backend/hook.js
+++ b/src/backend/hook.js
@@ -21,9 +21,10 @@ export function installHook (window) {
     },
 
     once (event, fn) {
+      const event_alias = event;
       event = '$' + event
       function on () {
-        this.off(event, on)
+        this.off(event_alias, on)
         fn.apply(this, arguments)
       }
       ;(listeners[event] || (listeners[event] = [])).push(on)


src/backend/hook.js
```diff
@@ -30,13 +30,15 @@ export function installHook (window) {
...
+      if (!/^\$/.test(event)) {
+	event = '$' + event
+      }
       if (!arguments.length) {
         listeners = {}
       } else {
         const cbs = listeners[event]
         if (cbs) {
-          if (!fn) {
+          if (!fn || cbs.length === 1) {
```
Akryum: @maxushuang Why are you making this change?

posva:
Hey sorry for the delay and thanks for the PR!
Could you explain a bit more what is not working as expected right now and is fixed by your PR?

maxushuang:
Hey, sorry, reply delay. 
The vue-devtools source in src/backend/hook.js is below.
```
const hook = {
    once (event, fn) {
      event = '$' + event
      function on () {
        this.off(event, on)
        fn.apply(this, arguments)
      }
      ;(listeners[event] || (listeners[event] = [])).push(on)
    },
   off (event, fn) {
      event = '$' + event
      if (!arguments.length) {
        listeners = {}
      } else {
     ......
};
```

```
hook.once('event1', _ => {});

hook.emit('event1');
```
### 

When hook emits event1, hook will call ```this.off(event, on)```. But, now, the event in this.off has becomes ``` $event1 ```, hook will add a '$' symbol again when call off function. 
So hook will not off event correctly.

PR #474, opened 2017-12-19
Support native Date type, Components, improved RegExp, new CustomValue API

### LINKED ISSUE #460: Property or method "toJSON" is not defined
This error happens when i pass the component by param to action in vuex.

To ***reproduce*** the error
```javascript
  //  Call an action in component passing the component as param to vuex
  this.$store.dispatch('metadataAddBox', { component: this });
```

screenshot

> [Vue warn]: Property or method "toJSON" is not defined on the instance but referenced during render. Make sure to declare reactive data properties in the data option.

### LINKED ISSUE #475: Dates are restored as Strings when time traveling
Using time-travel debugging with a store that contains date will lead to errors like this:

```
Error in render: "TypeError: b.date.getTime is not a function"
```

This is because the dates are restored as Strings instead of actual `Date` objects.

Akryum:
Fix #475 Fix #460

This PR improves `Date` and components support:
- Dates are now treated as native (like RegExp) and are fomated as such.
- Dates and reg exps are now correctly revived when time traveling in vuex.
- Import/Export of dates and reg exps now works too.

image

- New `_custom` API.

Example for components:

```js
{
    _custom: {
      type: 'component',
      id: instance.__VUE_DEVTOOLS_UID__,
      display: getInstanceName(instance),
      value: { /* component state */ },
      fields: {
        abstract: true
      }
    }
  }
```

Other example: https://gist.github.com/Akryum/f2e08df47088d5be0f933b67e90261dc

- Components are now handled in a specific way using above new feature:

capture d ecran du 2017-12-25 18 52 28

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -21,7 +21,7 @@ const hook = window.__VUE_DEVTOOLS_GLOBAL_HOOK__
 const rootInstances = []
 const propModes = ['default', 'sync', 'once']
 
-const instanceMap = window.__VUE_DEVTOOLS_INSTANCE_MAP__ = new Map()
+export const instanceMap = window.__VUE_DEVTOOLS_INSTANCE_MAP__ = new Map()
 const consoleBoundInstances = Array(5)
 let currentInspectedId
 let bridge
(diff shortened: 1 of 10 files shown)


src/util.js
```diff
@@ -43,7 +43,8 @@ export function stringify (data) {
   return CircularJSON.stringify(data, replacer)
 }
 
-function replacer (key, val) {
+function replacer (key) {
+  const val = this[key]
```
Akryum: @yyx990803 This is to prevent the default replacer to turn Dates into Strings before calling the custom replacer. See https://github.com/WebReflection/circular-json/issues/4

Akryum:
@posva Ready for review! :smile:

Akryum:
I added a new `_custom`  object API to be able to send complex customizable data to the devtools state inspector. This currently allows displaying components in a specific way. See OP.

Akryum:
I have made some changes on #486 to make the CustomValue API much more powerful. For example: https://gist.github.com/Akryum/f2e08df47088d5be0f933b67e90261dc

src/backend/index.js
```diff
@@ -320,18 +320,46 @@ function getInstanceDetails (id) {
...
+      type: 'component',
+      id: instance.__VUE_DEVTOOLS_UID__,
+      display: getInstanceName(instance),
+      state: reduceStateList(state)
+    }
+  }
+}
+
+export function reduceStateList (list) {
+  if (!list.length) {
```
posva: minor, but you can write this as `if (!list.length) return`

PR #476, opened 2017-12-20
Select component from dom tree

nicoeg:
Inspired by the react developer tools and of course Chromes developer tools.
This PR adds functionality for selecting a component from the dom tree. 

Instead of just highlighting elements it highlights components themselves if the hovered element is not a Vue component it'll select the parent component.
The component tree will expand if the selected component is minimized.

dec-20-2017 22-39-20

Edit:
Just saw Zzuligy's [PR](https://github.com/vuejs/vue-devtools/pull/469). 
While I think that its cool that he's using the developer tools selector I still feel that our PR's have some major differences. Eg. that my version will allow the feature to also function in Safari and Firefox. 
I've messages Zzuligy to hear how he feels.

### DIFF
--- a/src/backend/highlighter.js
+++ b/src/backend/highlighter.js
@@ -1,10 +1,23 @@
 import { inDoc } from '../util'
+import { getInstanceName } from './index'
 
 const overlay = document.createElement('div')
 overlay.style.backgroundColor = 'rgba(104, 182, 255, 0.35)'
 overlay.style.position = 'fixed'
 overlay.style.zIndex = '99999999999999'
 overlay.style.pointerEvents = 'none'
+overlay.style.display = 'flex'
+overlay.style.alignItems = 'center'
+overlay.style.justifyContent = 'center'
+overlay.style.borderRadius = '3px'
+const overlayContent = document.createElement('div')
+overlayContent.style.backgroundColor = 'rgba(104, 182, 255, 0.9)'
+overlayContent.style.fontFamily = 'monospace'
+overlayContent.style.fontSize = '11px'
+overlayContent.style.padding = '2px 3px'
+overlayContent.style.borderRadius = '3px'
+overlayContent.style.color = 'white'
+overlay.appendChild(overlayContent)
 
 /**
  * Highlight an instance.
@@ -16,7 +29,10 @@ export function highlight (instance) {
   if (!instance) return
   const rect = getInstanceRect(instance)
   if (rect) {
-    showOverlay(rect)
+    let content = ''
+    const name = getInstanceName(instance)
+    name && (content = `<span style="opacity: .6;">&lt;</span>${name}<span style="opacity: .6;">&gt;</span>`)
+    showOverlay(rect, content)
   }
 }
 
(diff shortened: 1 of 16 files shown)


nicoeg:
@Akryum 
Yeah, I just thought since both our PR's are doing some of the same stuff like toggling all parent instances that we could agree on a way to do it so we don't have multiple methods in the codebase.

nicoeg:
@Akryum I see you've added a better way to toggle parent instances in #479. Should I update this pr to follow that approach once yours is merged?

nicoeg:
@Akryum Should be ready to review now

Rebased, updated to use new `inspect-instance` method and added tooltip to the  button. 
screen shot 2018-01-13 at 12 39 07

Akryum:
vue-devtools-highlighter

yyx990803:
Looks good to me!

posva:
I thought the right click extra item was enough, but displaying the name of the component is also really nice!

Akryum:
> I thought the right click extra item was enough
Well, the browser offer both options :smile:

src/backend/highlighter.js
```diff
@@ -16,7 +29,10 @@ export function highlight (instance) {
   if (!instance) return
   const rect = getInstanceRect(instance)
   if (rect) {
-    showOverlay(rect)
+    let content = ''
+    const name = getInstanceName(instance)
+    name && (content = `<span style="opacity: .6;">&lt;</span>${name}<span style="opacity: .6;">&gt;</span>`)
```
posva: I would replace this with

```js
if (name) {
	content = ...
	showOverlay(...)
}
```
since it doesn't look like it's displaying anything otherwise

src/backend/utils.js
```diff
@@ -0,0 +1,6 @@
+export function searchComponent (el) {
```
posva: I think we could have a more explicit name like `findRootComponent`

Akryum: We don't find the root component, we find the closest component that contains the element.

posva: yes, sorry, I updated the comment right away 😆

Akryum:
@yyx990803 I think I'm done with this PR 😄

Akryum:
vue-devtools-select-tooltip

src/backend/highlighter.js
Akryum: It can still show the component boundaries even if the component doesn't have a name (most certainly using an old version of vue-loader or vue).

src/devtools/mixins/keyboard.js
```diff
@@ -1,23 +1,20 @@
...
-  37: 'left',
-  38: 'up',
-  39: 'right',
-  40: 'down'
-}
+export const LEFT = 37
+export const UP = 38
+export const RIGHT = 39
+export const DOWN = 40
+export const S = 83
```
posva: Do you also type in Dvorak? 😆
Using Keycodes doesn't work across keyboards for letters

Akryum: Why? I just tried in Dvorak to be sure and it's working fine... :confused:

posva: You're completely right. I had recently a problem with keycodes but it's something a bit different, it only happens with the digits row

PR #477, opened 2017-12-23
Collapsable and redesigned state inspector

Akryum:
- Sections in state inspector are now collapsable
- Shift + Click and Ctrl + Click to expand/collapse all sections
- They are also redesigned
- Some style refactoring
- Vuex history items have now the same height as Events history items

vue-devtool-collpase-data-sections

### DIFF
--- a/src/devtools/components/DataField.vue
+++ b/src/devtools/components/DataField.vue
@@ -1,16 +1,19 @@
 <template>
   <div class="data-field">
-    <div class="self"
+    <div
+      class="self"
+      :style="{ marginLeft: depth * 14 + 'px' }"
       @click="toggle"
-      :style="{ marginLeft: depth * 14 + 'px' }">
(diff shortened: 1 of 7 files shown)


posva:
Looks nice, I prefer it to #392

posva:
Testing on mac, it doesn't seem to expand or collapse all. Also, I would update the popover to display cmd or the unicode symbol

Akryum:
Fixed by https://github.com/Akryum/vue-devtools/commit/5b7745d5d79f7e12341f5072d4ad3f2061c59ae2

posva:
Works nicely now!

Some extra thoughts for the future:

- remove the pointer cursor on properties (something like `.key:not(.abstract), .colon, .value {cursor: initial;}`
- make some of the tree parts non selectable (this may have some edge cases

PR #478, opened 2017-12-23
Scroll to component button

### LINKED ISSUE #235: Disable components pane auto-scroll
In the standard Chrome dev-tools, a user must right-click on an element in the elements pane and select scroll into view.  In the Vuejs Chrome Dev-tools, the root element of a component is scrolled to when a component in the components pane is clicked on.  Sometimes when viewing a large (large height, or large width) component and looking at a child, when I click on the component in the components pane, the window scrolls which is not what I want while inspecting the child.

My request is that there be right-click menu to behave how the elements pane does, or that there is an option to disable auto-scroll. I can start work on this, but want to know whether either of these changes are acceptable.

[image](https://umaar.com/dev-tips/20-reveal-in-elements/)

Akryum:
Fix #235 

- Scroll to component button in Component tree
- Center the instance to the middle of the screen
- Added some routing to target page

vue-devtools-scroll-instance

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -591,7 +591,7 @@ function processObservables (instance) {
 function scrollIntoView (instance) {
   const rect = getInstanceRect(instance)
   if (rect) {
-    window.scrollBy(0, rect.top)
+    window.scrollBy(0, rect.top + (rect.height - window.innerHeight) / 2)
   }
 }
 
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -63,12 +63,16 @@ function connect () {
     currentInspectedId = id
     const instance = instanceMap.get(id)
     if (instance) {
-      scrollIntoView(instance)
       highlight(instance)
     }
     bindToConsole(instance)
     flush()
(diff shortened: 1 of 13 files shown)


src/backend/index.js
```diff
@@ -587,7 +591,7 @@ function processObservables (instance) {
 function scrollIntoView (instance) {
   const rect = getInstanceRect(instance)
   if (rect) {
-    window.scrollBy(0, rect.top)
+    window.scrollBy(0, rect.top + (rect.height - window.innerHeight) / 2)
```
posva: making it appear in the middle of the screen is pretty neat actually, It did bothered me some times (toolbars haha) when it scrolls to the very top

Akryum: > Not sure how to change this though (and I feel like it's a new change?)

What do you mean? 😕

davestewart: We have a component that's often taller then the screen, so when it scrolls to the middle, it's a bit odd. It would be good to scroll to middle for smaller components, but scroll to top for components taller than the screen height.

posva: Or making sure we never scroll more than half of the screen height

PR #479, opened 2017-12-24
Right click Inspect component + this.$inspect()

### LINKED ISSUE #242: Right click to inspect component
The feature I'm requesting is a context menu button to inspect a component. It can be annoying having to walk down your component tree to get to the one you want to inspect. We should be able to set up a context menu to walk up the dom tree until it finds an element that is being managed by our Vue app. If this feature is acceptable, I'll take a crack at it.

Akryum:
Fix #242 

- Right-click inspect
- `this.$inspect()`

vue-devtools-right-click-inspect

### DIFF
--- a/shells/chrome/src/devtools-background.js
+++ b/shells/chrome/src/devtools-background.js
@@ -76,12 +76,12 @@ function onContextMenu (info, tab) {
 // Action that may execute immediatly
 // or later when the Vue panel is ready
 
-function panelAction (cb, message) {
+function panelAction (cb, message = null) {
   if (created && panelShown) {
     cb()
   } else {
     pendingAction = cb
-    toast(message)
+    message && toast(message)
   }
 }
 
--- a/shells/chrome/src/devtools-background.js
+++ b/shells/chrome/src/devtools-background.js
@@ -3,6 +3,8 @@
 // Vue presence on the page. If yes, create the Vue panel; otherwise poll
 // for 10 seconds.
 
+let panelShown = false
+let pendingAction
 let created = false
 let checkCount = 0
 
@@ -14,6 +16,7 @@ function createPanelIfHasVue () {
   if (created || checkCount++ > 10) {
     return
   }
+  panelShown = false
   chrome.devtools.inspectedWindow.eval(
     '!!(window.__VUE_DEVTOOLS_GLOBAL_HOOK__.Vue)',
     function (hasVue) {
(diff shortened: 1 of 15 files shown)


posva:
It works, good job!
Is it possible to disable or remove the option if the right click isn't on a vue component (outside of Vue scope). I was testing on Vue docs and right-clicking outside of examples gives no feedback

shells/chrome/src/devtools-background.js
```diff
@@ -24,10 +27,65 @@ function createPanelIfHasVue () {
...
+// Execute pending action when Vue panel is ready
+
+chrome.runtime.onMessage.addListener(request => {
+  if (request === 'vue-panel-load') {
+    onPanelLoad()
+  }
+})
+
+function onPanelLoad () {
+  pendingAction && pendingAction()
```
posva: when does the pendingAction gets executed? I couldn't test this one
Depending on when it is triggered, it may be better to set `pendingAction` back to `null`

posva:
Some other thoughts:

- if possible, it would be nice to open devtools and focus vue-devtools as well
- did you get any chance of testing this on Firefox? I couldn't load a plugin from a file, I think I need to package it first

Akryum:
> Is it possible to disable or remove the option if the right click isn't on a vue component (outside of Vue scope).

Maybe. I'll investigate this. At least we could display something like a toast with a message.

> if possible, it would be nice to open devtools and focus vue-devtools as well

Unfortunately, this is impossible. Devtools must be open manually.

> did you get any chance of testing this on Firefox?

No, will try.

shells/chrome/src/devtools-background.js
Akryum: Fixed

Akryum:
vue-devtools-toast

Akryum:
vue-devtools-toast2

Akryum:
vue-devtools-toast3

Akryum:
capture d ecran du 2017-12-24 17 28 54

zzuligy:
nice work.put the inspect in menu could make the workflow fast which is better than react devtool.
but when i run,there is a error.
**`Unchecked runtime.lastError while running contextMenus.create: Cannot create item with duplicate id vue-inspect-instance`**

Akryum:
@zzuligy Fixed! :wrench:

Akryum:
@posva I added `this.$inspect()` in this PR as well, since it was mostly the same need.

Akryum:
Tested in Firefox :ok_hand:

scottbedard:
Thank you for this!

pSchaub:
Can we merge this now? This feature is really awesome! :+1:

PR #483, opened 2017-12-26
open-in-editor

Akryum:
- Component file tooltip
- Open in editor on click

For this to work, user webpack config needs to be modified. Example: https://github.com/vuejs/vue-devtools/pull/483/files#diff-b4cdb7abeacb810177ac96ccfd26094a ([supported editors](https://github.com/lahmatiy/open-in-editor#editor))

This should work with Nuxt out-of-the-box (using the same `/_open` route).


package.json
```diff
@@ -39,6 +39,7 @@
     "css-loader": "^0.28.7",
     "eslint": "^4.13.1",
     "eslint-plugin-vue-libs": "^2.0.1",
+    "express-open-in-editor": "^3.1.1",
```
yyx990803: why are we using the express middleware instead of `open-in-editor` itself?

Akryum: It's to be used in the webpack-dev-server directly (that uses an express server).

Akryum: See https://github.com/lahmatiy/express-open-in-editor#using-with-webpack-dev-server

yyx990803:
Hmm, this would not work when installed as an extension, since there will be no devServer running...

Akryum:
It needs some change to the user webpack config yes. We can't open the file from the extension anyway.

LinusBorg:
I don't think that's a downer, that's reasonable and can be documented.

I would include it in the webpack template asap. :)

src/devtools/views/components/ComponentInspector.vue
```diff
@@ -74,7 +74,23 @@ export default {
       } else {
         window.alert('DOM inspection is not supported in this shell.')
       }
+    },
+    onTitleClick () {
+      const file = this.target.file
+      if (file) {
+        fetch(`/_open?file=${file}`).then(() => {
```
yyx990803: So this will be making a request from the origin of the extension and only works during dev - however we should be able to initiate the request from the inspected window by doing an eval like in `inspectDOM`!

anteriovieira:
> This should work with Nuxt out-of-the-box (using the same /_open route).

I've done some testing and the implementation with nuxt is really very simple.

`npm install express-open-in-editor`

```js
// nuxt.config.js
const openInEditor = require('express-open-in-editor')

module.exports = {
  ...
  serverMiddleware: [
    {
      path: '/_open',
      handler: openInEditor({
        editor: 'code',
      })
    }
  ]
}
```

or if you want to use something less verbose you can install the [nuxt-open-in-editor](https://github.com/anteriovieira/nuxt-open-in-editor) module.

`npm install nuxt-open-in-editor`

```js
// nuxt.config.js
module.exports = {
  ...
  modules: [
    'nuxt-open-in-editor'
  ]
}
```

Akryum:
@atinux Shouldn't this work out-of-the-box with https://github.com/nuxt/nuxt.js/blob/9b8cbed51236d32372706be7c2381df032c3dd88/lib/core/renderer.js#L227-L249?

Atinux:
Actually it's @pi0 whom implemented it, but yeah it should work out-of-the-box with Nuxt.js directly.

pi0:
Basically, Nuxt should also work as our middleware is using exactly same query param (`?file=`)

If it's going to be a standard endpoint, I would suggest using something like `/open-in-editor` or `'/__open-in-editor' to prevent route conflicts. We can deliver it for 1.0.0 if you agree :)

yyx990803:
The new `vue-cli` will also have this built-in, so yeah let's standardize this.

`/__open-in-editor` sounds good to me.

anteriovieira:
`/__open-in-editor` sounds good to me too.

Akryum:
Updated to `/__open-in-editor` in #486

PR #489, opened 2018-01-05
Updated installation commands with production arguments

ematipico:
In order to install the extension manually, only the `dependencies` are needed. So, I took the liberty to add the `--production` flag to the instructions

```diff
-- npm install
++ npm install --production
```

```diff
-- yarn install
++ yarn install --prod
```

(`--prod` is an alias for `--production`)


Akryum:
The `npm run build` needs the devDependencies like webpack and vue-loader to work.

PR #491, opened 2018-01-08
show Vuex mutations commited before devtools init

### LINKED ISSUE #408: Vue Devtools not capturing initial vuex commits
Hi.

I have a simple application which fetches users on initialization. This uses the `FETCH_USERS` action and commits the mutation `USERS_FETCHED` when done. I never see this commit in the devtools, presumably because the devtools are not ready yet.

In the example below I use setTimeout to wait 500ms. Then the commit always shows in the devtools. 

It would be very nice if this could be avoided. Am I doing something wrong? The same problem is referenced here: https://github.com/vuejs/vue-devtools/issues/124 however the issue was closed with the message that the devtools should be open, then the page should be refreshed. But it doesn't solve the problem.

This is my `main.js`, which dispatches the action that doesn't show up as a commit:

```js
import Vue from "vue";
import VueRouter from "vue-router";
import Vuex from "vuex";

import { FETCH_USERS } from "./store";

import styles from "./styles.css";
import routes from "./routes";

import Application from "./components/Application";
import store from "./store";

Vue.use(VueRouter);

const router = new VueRouter({
  mode: "history",
  routes
});

new Vue({
  el: "#app",
  store,
  render: h => h(Application),
  router,
  created() {
    this.$store.dispatch(FETCH_USERS);
  }
});
```

if I change the initialisation to the following, the initial event is captured.

```js

setTimeout(
  () =>
    new Vue({
      el: "#app",
      store,
      render: h => h(Application),
      router,
      created() {
        this.$store.dispatch(FETCH_USERS);
      }
    }),
  500
);
```

Can I fix this in some way?

Oh, and by the way, I'm running Chrome 61.

Thanks for your great work!

screenshot from 2017-09-11 13-51-53

---

repro:

```html
<html>
<head>
  <title>Welcome to Vue</title>
  <script src="https://unpkg.com/vuex"></script>
</head>
<body>
  <div id="app">
  </div>

  <script>
  function webRequest() {
    return {username: 'vuejs-hackman'};
  }
  var store = new Vuex.Store({
    state: {
      user: {}
    },
    mutations: {
      STORE_DATA(state, payload) {
        console.log('irun');
        window.localStorage.setItem("user", JSON.stringify(payload));
        state.user = payload;
      }
    },
    actions: {
      loadUser: function (ctx, payload) {
        ctx.commit('STORE_DATA', payload)
      }
    }
  });

  var app = new Vue({
    el: '#app',
    store: store,
    methods: {
      initUser: function(user) {
        // this.$store.commit('STORE_DATA', user);
        this.$store.dispatch('loadUser', webRequest());
      },
      loadUser: function() {
        // this.$store.dispatch('loadUser', webRequest());
        this.$store.commit('STORE_DATA', {user:'hi'});
      }
    },
    template: '<button @click="loadUser">click</button>',
    created: function() {
      const user = window.localStorage.getItem("user");
      if (user) {
        this.initUser(JSON.parse(user));
      }
    }
  })
  </script>
</body>
</html>
```

bartlomieju:
This PR provides proof-of-concept for displaying all of Vuex store. 

I am not really happy with this code as it relies on setting store._devtoolsBuffer = undefined after vue-devtools boots up in order to prevent memory leak. 

Anyway, thanks to this solution all mutations are inspectable in devtools.

Related to vuejs/vuex#1128.

Closes #408 
  
I'll be happy to get some feedback


bartlomieju:
I've added additional property `supportsVuexMutation` to global devtools hook to prevent memory leaks in Vuex library in case user uses old version of vue-devtools.

Akryum:
Looks good to me, waiting vuejs/vuex#1128

simplesmiler:
There is an issue. State snapshots are made at the moment of "replaying" of the buffer instead of during recording, which makes almost all of them (including the "base state") incorrect.

I managed to fix this by exposing `getSnapshot(store)` through the hook, and saving snapshots alongside mutations on the vuex side until devtools kicks in and extracts them.

To avoid creating additional pull requests, here are fixes:
- https://github.com/simplesmiler/vue-devtools/commits/vuex-mutation-buffer
- https://github.com/simplesmiler/vuex/commits/devtools-buffer

bartlomieju:
#723 has landed in beta so we don't need it anymore

PR #506, opened 2018-01-13
499 Invalid prop definition crash

### LINKED ISSUE #499: Cannot read property 'type' of null
OS : Ubuntu 16.04
Chrome : 63.0.3239.84
vue : 2.5.13
vuex : 3.0.1
vue-devtools: 4.0.0

No content is appeared at right side, when I click on a component to inspect it. 
After I click on a component to inspect it, there are also a lot of errors in my console.

Please look at these two images below.

image

image

Akryum:
Fix #499

### DIFF
--- a/shells/dev/target/Page1.vue
+++ b/shells/dev/target/Page1.vue
@@ -1,3 +1,13 @@
 <template>
-  <h2>Page 1</h2>
+  <div id="page-1">
+    <h2>Page 1</h2>
+    <p>{{ provided }}</p>
+  </div>
 </template>
+
+<script>
+export default {
+  inject: ['provided']
+}
+</script>
+
--- a/shells/dev/target/Router.vue
+++ b/shells/dev/target/Router.vue
@@ -11,6 +11,15 @@
   </div>
 </template>
 
+<script>
+export default {
+  provide: {
+    provided: 'foo'
+  }
+}
+</script>
+
+
 <style scoped>
 a {
   margin-right: 6px;
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -455,10 +455,10 @@ function processProps (instance) {
         type: 'props',
         key,
         value: instance[key],
-        meta: {
+        meta: prop ? {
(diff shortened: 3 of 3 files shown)


Akryum:
It looks like vee-validate is doing quite weird things with props. I added a guard against invalid prop definitions. This is so edgy I don't know if it's really the devtools that should have fix for this or vee-validate. Thoughts @yyx990803?

Akryum:
@logaretm ⬆️

Akryum:
Also, the issue doesn't seems related to provide/inject.

logaretm:
I can provide a fix on my end. But what is considered an invalid prop? vee-validate adds mostly adds data/computed fields to the components in the `beforeCreated` hook.

And there is the `fields` object which is consistently being updated since the props depend directly on the vnodes currently present in the template.

logaretm:
@Akryum My guess is correct. The `fields` property seems to be the culprit here:

```js
this.$options.computed[options.fieldsBagName || 'fields'] = function fieldBagGetter () {
  if (!Object.keys(this.$validator.flags).length) {
    return fakeFlags;
  }
  return this.$validator.flags;
};
```
The check `Object.keys(this.$validator.flags).length` provides a temporary fallback for the fields until they are present after the directives have been bound to prevent undefined or null access crashes.

Removing the check fixes the issue in my local tests. I can push a release quickly and I would appreciate it if you can point out violations that are being done by vee-validate so I can address them too.

Akryum:
From what I understand, `fields` looks like a Vue component and the devtools are now displaying components in the inspector. But the `$options.props` object of `fields` is invalid (each prop definition minimal form is `myProp: { type: null }`.

Akryum:
Also, the name resolved for it is something like `<{ "dirty": false, "valid": true, ... }>`.

logaretm:
But it isn't a component nor a prop, is it? so to clarify It just looks like one and gets interpreted as one and this caused the issue.

Akryum:
We check if `object._isVue` is `true` and we treat it as a component then.

PR #509, opened 2018-01-14
501 - Encode file path as URI + fix console display on Windows

### LINKED ISSUE #501: Open in editor path is missing slashes
When opening file from devtools there are missing slashes in path for file. This issue is caused because of template literals is escaping with `\`.

obrazok

To fast fix this issue simply add replace on this line: https://github.com/vuejs/vue-devtools/blob/24812e5542bb359a06a3e62379fd8c8633ae6034/src/devtools/views/components/ComponentInspector.vue#L87
`const file = this.target.file.replace(/\\/g, '/')`

Akryum:
Fix #501

### DIFF
--- a/src/devtools/views/components/ComponentInspector.vue
+++ b/src/devtools/views/components/ComponentInspector.vue
@@ -101,27 +101,25 @@ export default {
       const file = this.target.file
       // Console display
       const fileName = file.replace(/\\/g, '\\\\')
-      if (file) {
-        const src = `fetch('/__open-in-editor?file=${encodeURI(file)}').then(response => {
-          if (response.ok) {
-            console.log('File ${fileName} opened in editor')
+      const src = `fetch('/__open-in-editor?file=${encodeURI(file)}').then(response => {
+        if (response.ok) {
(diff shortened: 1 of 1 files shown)


src/devtools/views/components/ComponentInspector.vue
```diff
@@ -99,18 +99,21 @@ export default {
...
-          const msg = 'Opening component ${file} failed'
+          const msg = 'Opening component ${fileName} failed'
           if (__VUE_DEVTOOLS_TOAST__) {
             __VUE_DEVTOOLS_TOAST__(msg, 'error')
           } else {
             console.log('%c' + msg, 'color:red')
           }
           console.log('Check the setup of your project, see https://github.com/vuejs/vue-devtools/blob/master/docs/open-in-editor.md')
         }
+        console.log('Check the setup of your project, see https://github.com/vuejs/vue-devtools/blob/master/docs/open-in-editor.md')
```
michalsnik: Did you put the same console.log on purpose here? In case the response is not ok you'll get 2 same logs in the console.

PR #519, opened 2018-01-15
New keyboard shortcuts + tooltips refactoring

Akryum:
- [New keyboard shortcuts](https://docs.google.com/spreadsheets/d/1iv3cBE1BpW8mRZLZQDWDbXMmncscmplZ6tJaYocue48/edit?usp=sharing)
- Teaks for shortcuts in tooltips
- New `$t` translation plugin
- Tooltips with shortcuts are now declared in a locale file (easier to manage)
- Improved entry lists keyboard navigation (scrolling)

We may want to move all texts to the locale file. Also, this open up translation for a future PR.


src/devtools/views/events/module.js
```diff
@@ -24,6 +24,8 @@ const mutations = {
     state.inspectedIndex = -1
   },
   'INSPECT' (state, index) {
+    if (index < 0) index = 0
+    if (index >= state.events.length) index = state.events.length - 1
```
michalsnik: wouldn't this be enough actually?
```
state.inspectedIndex = Math.max(0, state.events.length - 1)
```

src/devtools/plugins.js
```diff
@@ -8,3 +10,19 @@ Vue.use(VTooltip, {
...
+const locales = require.context('./locales')
+const keyboardReg = /\[(\w+)\]/g
+const iconReg = /\|(\w+)\|/g
+Vue.use(VI18n, {
+  strings: locales(`./${currentLocale}`).default,
+  defaultValues: {
+    keys
+  },
+  replacer: text => {
+    text = text.replace(keyboardReg, '<span class="keyboard">$1</span>')
```
michalsnik: I don't like this solution. Now whenever `[]` or `||` appears somewhere in translation, it's going to wrap this part with one of those elements. As for the icon - if it's a generic app-wide solution I'm ok with it, but probably with better interpolation signs. e.g `<<iconName>>`, or at least `||iconName||` - though the first one more readable I guess.
But keyboard is rather case specific thing, and I wouldn't put it here. I'm thinking about alternative ways, and perhaps it would be better to make a dedicated file just for tooltips content, and import keys explicitly there.
That would leave us with translations in tact, without extra concerns.

Akryum: We are going to have a lot of keyboard shortcuts in the strings--more than icons 😛

src/devtools/views/events/module.js
Akryum: No? 🙃

Akryum:
Here is a summary of the keyboard shortcuts: https://docs.google.com/spreadsheets/d/1iv3cBE1BpW8mRZLZQDWDbXMmncscmplZ6tJaYocue48/edit?usp=sharing

Akryum:
The order of modifiers is:
1. `ctrl` (also works with Mac `meta` key)
2. `shift`
3. `alt`

Examples of valid modifiers:

```js
'' // No modifier
'ctrl'
'ctrl+shift'
'ctrl+alt'
'ctrl+shift+alt'
```

Akryum:
@posva Changed to Ctrl|Meta + 1..3

src/devtools/views/events/module.js
michalsnik: Ah, true. It would have to be:
```
Math.min(Math.max(0, index), state.events.length - 1)
```
but it wouldn't be necessarily more readable ;D

src/devtools/mixins/entry-list.js
```diff
@@ -0,0 +1,12 @@
+import { scrollIntoView } from 'src/util'
+
+export default {
+  watch: {
+    inspectedIndex (value) {
+      this.$nextTick(() => {
+        const el = value === -1 ? this.$refs.baseEntry : this.$refs.entries[value]
+        el && scrollIntoView(document.querySelector('.left .scroll'), el, false)
```
michalsnik: I recommend we use `.js-` prefixed classes for everything that is being handled by JS. It would allow us to more safely deal with the code and potential refactors.

posva:
Ha, realized later that Cmd+numbers may not be that good because `cmd+shift+3` does a screenshot in Mac so if your number is on shift, you can use it. I tested on programmer dvorak and none work. Maybe we could use `alt+{` and `alt+}` to change tabs, a bit like `cmd+{`/`ctrl+{` (that I think do not work on azerty keyboard because you have to use shift as well 😆 )

On another topic, I think the `keydown` isn't the most explicit solution to bind to shortcuts, it's not because I created it 😆  but I think https://github.com/shentao/vue-global-events could be helpful. I added a feature (still on PR) to filter events (could be used to not trigger events on inputs and textareas as you did)

Akryum:
> Ha, realized later that Cmd+numbers may not be that good because cmd+shift+3 does a screenshot in Mac so if your number is on shift, you can use it. I tested on programmer dvorak and none work.

Tested and working even with Dvorak layout. Just don't press Shift key :stuck_out_tongue:

Issue #515, opened 2018-01-15
Message length exceeded maximum allowed length


w1ndy:
### Version
4.0.0

### Browser and OS info
Chrome 63

### Steps to reproduce
Put a large amount of data in components and open Vue devtools.

### What is expected?
Devtools open correctly.

### What is actually happening?
![vue-devtools](https://user-images.githubusercontent.com/652068/34939784-0c9d1cd8-fa28-11e7-9764-6ba51406b81a.png)
Error thrown: Message length exceeded maximum allowed length.
Could be similar to [this](https://github.com/BabylonJS/Spector.js/issues/69).

w1ndy:
A minimal repo:
```html
<!DOCTYPE html>
<body>
    <div id="app"></div>
</body>
<script src="https://cdnjs.cloudflare.com/ajax/libs/vue/2.5.13/vue.js"></script>
<script>
new Vue({
    el: '#app',
    mounted () {
        for (let i = 0; i < 10000000; i++)
            this.largeArray.push(i)
    },
    data: {
        largeArray: []
    }
})
</script>
```

w1ndy:
vue-devtools completely freezes on this code, and the error will be thrown a moment later.

Akryum:
Please try this branch: https://github.com/Akryum/vue-devtools/tree/component-inspect-optimizations

Akryum:
Please try this branch: https://github.com/Akryum/vue-devtools/tree/component-inspect-optimizations

w1ndy:
@Akryum Sorry for the late reply, but the patch merged did not resolve this issue... The latest vue-devtools is still throwing errors and violations on the example I gave earlier.

w1ndy:
Is it possible to not serialize and transfer data entirely, but retrieve their values on demand?

colinfike:
I can confirm that I am also still seeing this issue.

maverickchangithub:
@colinfike    I got the same issue in a PC environment  , but it is OK in a MAC dev machine.  Don't know why , same code , internet problem?

Akryum:
How much RAM do you have?

colinfike:
I don't actually have the machine I was working on at the time but I believe I had 16gb with 8gb allocated to docker.

maverickchangithub:
@Akryum now i have the same issue on a Mac dev machine too . I use 2048M ram in both machine , php 7.2 . Really strange issue  , maybe it is my code which broke the tool , not confirmed yet.

[...]

PR #523, opened 2018-01-16
Serialization & Types improvements

### LINKED ISSUE #133: When listing a Component's model, show the function signature of a value
If the value of a property of a Vue data model is a function. Vue Devtools currently just shows `"Function"` as the value. It would be neat if it showed the function signature instead (eg. `"function someFunction(arg)"`).

You can get the signature by calling `.toString()` on the function and getting everything up to the closing parenthesis:

```
if (typeof value === "function") {
    const valueStr = value.toString()
    return valueStr.slice(0, valueStr.indexOf(")") + 1)
}
```

### LINKED ISSUE #349: show structure Set
code like this
```javascript
data() {
   setKeys: new Set(),
}
```
vue-devtools will be show this
image
better cast **new Set()** to structure **Array** in show result

### LINKED ISSUE #499: Cannot read property 'type' of null

### LINKED ISSUE #507: v4.0.1 - "Uncaught TypeError: Cannot read property '_route' of undefined" Everywhere
On chrome devtools -> Vue, if I click on any component (on the component tree on the left side) this error appears on console:

`vue-router.esm.js?880d:557 Uncaught TypeError: Cannot read property '_route' of undefined
    at Vue$3.get (vue-router.esm.js?880d:557)
    at backend.js:1
    at x (backend.js:1)
    at t.a (backend.js:1)
    at Object.y (backend.js:1)
    at e (backend.js:1)
    at e (backend.js:1)
    at e (backend.js:1)
    at e (backend.js:1)
    at Object.t.stringifyStrict (backend.js:1)`

updated the the project packages, including vue-router, tried on an old version of the project that was working before, but with the v4.0.1 it gives the same error. Installed an old version of vue-devtools (3.1.7) and everything works well as before.

when events or other actions are triggered when using the website the console fills with errors:

On firefox it gives this error:
TypeError: this._routerRoot is undefined

### LINKED ISSUE #528: Uncaught RangeError: Maximum call stack size exceeded at Array.toString (native)
### Version
4.0.0

### Browser and OS info
Chrome 63.0.3239.132 (Official Build) (64-bit) / Windows 10

### Steps to reproduce
- Go to this [url](https://2wk5wvxrj.codesandbox.io/) to view the sample project on CodeSandbox.io
- Right-click the page and select "Inspect"
- In the Developer Tools window, click Vue menu to display Vue-Devtools
- Click on `<app>` on the left pane, the right pane should not cause any error
- Click on `<Vuetable>` node or its child component `<VuetableRowHeader>`, and it dumps error out in the console pane

### What is expected?
Vue-Devtools should display info about the selected component.

Before 4.0.0, this never happens.

The following error shows up in console pane

```
backend.js:1 Uncaught RangeError: Maximum call stack size exceeded
    at Array.toString (native)
    at D (chrome-extension://nhdogjmejiglipccpnnnanhbledajbpd/build/backend.js:1:20128)
...
```
 

---
The old version of Vue-Devtools (before 4.0) was working fine without any error showing up.

Akryum:
- CustomValue API changes
  - new `tooltip` option
  - fields order retained from `value`
- Special format (using CustomValue API) for:
  - Vuex Store
  - VueRouter
  - Component definition
  - Function (Fix #133)
  - Set & Map (Fix #349, Original PR #490)
- Serialization is more fault tolerant (Fix #499 Fix #507)
- Cache Custom values to prevent stack overflow (circular references, Fix #528)
- Switch the order of 'data' and 'props' sections in component inspector to better follow [the style guide](https://vuejs.org/v2/style-guide/#Component-instance-options-order-recommended)
- Performance improvements (related to #515) 

capture d ecran 2018-01-16 a 12 27 33
capture d ecran 2018-01-16 a 12 28 23
capture d ecran 2018-01-16 a 12 29 20

### DIFF
--- a/shells/dev/target/NativeTypes.vue
+++ b/shells/dev/target/NativeTypes.vue
@@ -18,6 +18,7 @@
 
 <script>
 import { mapState, mapGetters, mapMutations } from 'vuex'
+import CompDef from './Other.vue'
 
 export default {
   components: {
@@ -30,7 +31,18 @@ export default {
   data () {
     return {
       localDate: new Date(),
-      testComponent: null
+      reg: /abc/gi,
+      testComponent: null,
+      hello: function foo (a, b, c) {},
+      hey: function empty () {},
+      def: CompDef,
+      def2: {
+        name: 'MyComponent',
+        render () {}
+      },
+      def3: {
+        render () {}
+      }
     }
   },
   computed: {
--- a/shells/dev/target/NativeTypes.vue
+++ b/shells/dev/target/NativeTypes.vue
@@ -12,6 +12,7 @@
 
     <p>
       <button @click="sendComponent">Vuex mutation</button>
+      <button @click="createLargeArray">Create large array</button>
     </p>
   </div>
 </template>
@@ -31,6 +32,7 @@ export default {
       render: h => h('div', '<TestComponent />')
     }
   },
+
   data () {
     return {
       localDate: new Date(),
@@ -45,30 +47,45 @@ export default {
       },
       def3: {
         render () {}
-      }
+      },
+      largeArray: []
     }
   },
   computed: {
     ...mapState(['date']),
     ...mapGetters(['hours']),
+
     theRouter () {
       return this.$router
     },
+
     theStore () {
       return this.$store
     }
   },
+
   mounted () {
     this.testComponent = this.$refs.component
   },
+
   methods: {
     ...mapMutations({
       updateDate: 'UPDATE_DATE'
     }),
+
     sendComponent () {
       this.$store.commit('TEST_COMPONENT', this.testComponent)
+    },
+
+    createLargeArray () {
+      const list = []
+      for (let i = 0; i < 10000000; i++) {
+        list.push(i)
+      }
+      this.largeArray = list
     }
   },
+
   filters: {
     prototypeString: val => Object.prototype.toString.call(val)
   }
--- a/shells/dev/target/NativeTypes.vue
+++ b/shells/dev/target/NativeTypes.vue
@@ -11,8 +11,20 @@
     <TestComponent ref="component" />
 
     <p>
-      <button @click="sendComponent">Vuex mutation</button>
-      <button @click="createLargeArray">Create large array</button>
+      <button @click="sendComponent()">Vuex mutation</button>
+      <button @click="createLargeArray()">Create large array</button>
+    </p>
+
+    <h3>Set</h3>
+    <pre>{{ setDisplay() }}</pre>
+
+    <h3>Map</h3>
+    <pre>{{ mapDisplay() }}</pre>
+
+    <p>
+      <button @click="testVuexSet()">Vuex Set</button>
+      <button @click="testVuexMap()">Vuex Map</button>
+      <button @click="forceRefresh()">Refresh</button>
     </p>
   </div>
 </template>
@@ -54,8 +66,15 @@ export default {
     }
   },
   computed: {
-    ...mapState(['date']),
-    ...mapGetters(['hours']),
+    ...mapState([
+      'date',
+      'set',
+      'map'
+    ]),
+
+    ...mapGetters([
+      'hours'
+    ]),
 
     theRouter () {
       return this.$router
@@ -63,7 +82,7 @@ export default {
 
     theStore () {
       return this.$store
-    }
+    },
   },
 
   mounted () {
@@ -72,7 +91,9 @@ export default {
 
   methods: {
     ...mapMutations({
-      updateDate: 'UPDATE_DATE'
+      updateDate: 'UPDATE_DATE',
+      testVuexSet: 'TEST_SET',
+      testVuexMap: 'TEST_MAP'
     }),
 
     sendComponent () {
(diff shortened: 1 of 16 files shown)


Akryum:
@evan @posva @michalsnik I fixed #528 with a new `EncodeCache` class that should be used in front of any serialization involving CustomValue API. I also changed the order of data and props sections to better follow [the style guide](https://vuejs.org/v2/style-guide/#Component-instance-options-order-recommended). WDYT?

Akryum:
capture d ecran du 2018-01-19 12 45 30

Akryum:
I added some optimization from this branch: https://github.com/Akryum/vue-devtools/tree/component-inspect-optimizations ([changes](https://github.com/Akryum/vue-devtools/compare/a68ee61e07acf44d4ea7cb3b20a0e2493e734824...Akryum:component-inspect-optimizations))

x10 faster when selecting a component (tested with the `NativeTypes` component after clicking the `Create large array` button).

Akryum:
@yyx990803 I also made a PR https://github.com/yyx990803/circular-json-es6/pull/5 with the changes I made to CircularJSON.

Akryum:
I don't think we can handle WeakMap since they are not iterable.

src/util.js
```diff
@@ -52,29 +58,184 @@ export const SPECIAL_TOKENS = {
...
+    }
+  }
+}
+
+export function getCustomFunctionDetails (func) {
+  const args = func.toString().match(/\(.*\)/)[0]
+  return {
+    _custom: {
+      type: 'function',
+      display: `<span>${func.name}</span>${args}`
```
posva: I think adding the word _function_ somewhere could make this more explicit. What about something like _ƒ o(a, b, c)_, similar to the console?

posva:
Couldn't test open in editor yet, but everything else looks fine 🙂

Akryum:
capture d ecran du 2018-01-20 23 33 31

src/util.js
```diff
@@ -52,29 +58,184 @@ export const SPECIAL_TOKENS = {
...
-    return `[native Date ${val.toString()}]`
-  } else if (val && val._isVue) {
-    return getCustomInstanceDetails(val)
+  }
+  return sanitize(val)
+}
+
+export function getCustomMapDetails (val) {
+  const list = []
+  val.forEach(
```
michalsnik: You could use `map` instead:
```
const list = val.map((value, key) => ({ key, value}))
```

src/util.js
```diff
@@ -52,29 +58,184 @@ export const SPECIAL_TOKENS = {
...
+        abstract: true
+      }
+    }
+  }
+}
+
+export function reviveMap (val) {
+  const result = new Map()
+  const list = val._custom.value
+  for (let i = 0; i < list.length; i++) {
```
michalsnik: Perhaps you can use `forEach` instead? 
```
list.forEach(({ key, value}) =>
  result.set(key, reviver(null, value)
)
```

src/util.js
```diff
@@ -52,29 +58,184 @@ export const SPECIAL_TOKENS = {
...
+      value: list,
+      readOnly: true
+    }
+  }
+}
+
+export function reviveSet (val) {
+  const result = new Set()
+  const list = val._custom.value
+  for (let i = 0; i < list.length; i++) {
```
michalsnik: Same here

src/devtools/env.js
```diff
@@ -0,0 +1,23 @@
...
+  ctrl: isMac ? '&#8984;' : 'Ctrl',
+  shift: 'Shift',
+  alt: isMac ? '&#8997;' : 'Alt'
+}
+
+Object.defineProperties(Vue.prototype, {
+  '$isChrome': { get: () => isChrome },
+  '$isWindows': { get: () => isWindows },
+  '$isMac': { get: () => isMac },
+  '$isLinus': { get: () => isLinux },
```
michalsnik: Shouldn't it be `$isLinux`?

shells/dev/target/NativeTypes.vue
```diff
@@ -11,43 +11,116 @@
...
   },
+
   data () {
     return {
       localDate: new Date(),
-      testComponent: null
+      reg: /abc/gi,
+      testComponent: null,
+      hello: function foo (a, b, c) {},
+      hey: function empty () {},
```
michalsnik: Can you also add anonymous function here and make sure it works good as well?
`hi: function (a, b) {}` should be displayed probably as  `f (a, b)`, also what about arrow functions? Or `hi() {}`?

src/util.js
Akryum: `Map` doesn't have any `map` method.

src/util.js
Akryum: Simple loop without anonymous function call is faster imho. Also made a test case: https://jsperf.com/map-loop-vs-foreach (for what it's worth :upside_down_face:)

Akryum:
@michalsnik 
capture d ecran du 2018-01-22 11 24 10

Akryum:
@michalsnik 
capture d ecran du 2018-01-22 11 25 10

PR #533, opened 2018-01-19
Sync 'classifyComponents' settings with backend

Akryum:
Apply 'Format component names' user settings to component highlighter

### DIFF
--- a/src/backend/highlighter.js
+++ b/src/backend/highlighter.js
@@ -32,8 +32,8 @@ export function highlight (instance) {
   if (rect) {
     let content = ''
     let name = getInstanceName(instance)
-    claissifyComponents && (name = classify(name))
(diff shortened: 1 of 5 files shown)


src/backend/highlighter.js
```diff
@@ -30,7 +31,8 @@ export function highlight (instance) {
   const rect = getInstanceRect(instance)
   if (rect) {
     let content = ''
-    const name = getInstanceName(instance)
+    let name = getInstanceName(instance)
+    claissifyComponents && (name = classify(name))
```
posva: why are there so many `condition && statement` instead of `if (condition) statement` 😆 I think we should stop adding more of those

src/backend/index.js
```diff
@@ -39,6 +40,8 @@ export function initBackend (_bridge) {
   } else {
     hook.once('init', connect)
   }
+
+  bridge && config(bridge)
```
posva: when do you found yourself with no bridge?

PR #541, opened 2018-01-22
Dark mode color fixes

Akryum:
- Lightened abstract keyword in State inspector (was too dark)
- Change 'f' function from blue to purple (like numbers)


Akryum:
Before:

capture d ecran du 2018-01-22 17 33 08

After:

capture d ecran du 2018-01-22 17 32 48

PR #543, opened 2018-01-22
High-density mode

Akryum:
- Increase density of the State inspector if low height
- Automatically increase density of the component tree if there are too many rows

High density: tree
capture d ecran du 2018-01-22 18 14 48

High density: tree and state inspector
capture d ecran du 2018-01-22 18 15 44

High density: state inspector
capture d ecran du 2018-01-22 18 17 01


bartlomieju:
@Akryum maybe it's high time to introduce `Settings` tab? I guess there are a few elements users could tweak as they like.

yyx990803:
@bartlomieju good idea.

PR #556, opened 2018-01-24
555 - Fix HTML value not escaped in DataField

### LINKED ISSUE #555: HTML in strings get rendered as HTML in the component inspector instead of plain text
### Version
4.1.0

### Browser and OS info
Version 64.0.3282.119 (Official Build) (64-bit) / Windows 10

### Steps to reproduce
Inspect a component that has a string in its data with HTML

### What is expected?
Should see the string as HTML source code

Value is rendered as HTML

Akryum:
Fields of type 'String': formatted value is now escaped, fix #555

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -426,3 +426,18 @@ export function openInEditor (file) {
     eval(src)
   }
 }
+
+const ESC = {
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  '&': '&amp;'
+}
+
+export function escape (s) {
+  return s.replace(/[<>"&]/g, escapeChar)
+}
+
+function escapeChar (a) {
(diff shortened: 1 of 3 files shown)


src/util.js
```diff
@@ -426,3 +426,18 @@ export function openInEditor (file) {
     eval(src)
   }
 }
+
+const ESC = {
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
```
michalsnik: Don't we need to add `=` and `'` too?

Akryum:
Just copy-pasted the code from Vue here.

Le jeu. 25 janv. 2018 à 00:53, Michał Sajnóg <notifications@github.com> a
écrit :

PR #569, opened 2018-01-25
Use prototype toString to prevent bad serialization

### LINKED ISSUE #568: getCustomFunctionDetails throwing exception for anonymous functions
### Version
4.1.0

### Browser and OS info
Chrome 64.0.3282.119 / Windows 10

### Steps to reproduce
**This occurred on version 4.1.1 but that was not an option on the issue helper site**

Clone repository from: https://github.com/badams2527/devtools-issue

Install npm packages `npm install`

Run the project using `npm run dev`

Open the application in Chrome, then open Vue devtools.

In Vue devtools, navigate down the component tree to KendoGrid.

Click on KendoGrid.

Notice the console error

[Here's a screenshot of the error](https://imgur.com/a/TlNW3)

### What is expected?
It should populate the component instance details in the right panel. 

It is throwing an exception in the log and not populating the right panel

---
I set a debug point at that line in backend.js (after pretty printing), and checking the argument. It's returning
 
```
"function anonymous(d
/*``*/) {
return d.Total
}"
```

after the toString, leading me to believe it's just not set to handle anonymous functions properly.

simplesmiler:
Should fix #568

### DIFF
--- a/shells/dev/target/NativeTypes.vue
+++ b/shells/dev/target/NativeTypes.vue
@@ -33,6 +33,17 @@
 import { mapState, mapGetters, mapMutations } from 'vuex'
 import CompDef from './Other.vue'
 
+function setToString (func, string) {
+  return Object.defineProperty(func, 'toString', {
+    configurable: true,
+    enumerable: false,
+    value: () => string,
+    writable: true
+  })
+}
+
+const aWeirdFunction = setToString(function weird (a, b, c) {}, 'foo')
+
 export default {
   components: {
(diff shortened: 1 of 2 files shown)


simplesmiler:
Function that failed to be serialized properly will now show `(?)` as parameters.

Akryum:
Made it work anyway (but just in case, I left your check). 😄

simplesmiler:
👍 for `Function.prototype.toString.call`, didn't even cross my mind.

PR #572, opened 2018-01-26
Search improvements

### LINKED ISSUE #550: Crash while filtering a state with circular refs
### Version
4.1.0

### Browser and OS info
Chrome 63 / Linux

### Steps to reproduce
1. Inspect a component with circular references in the data
2. Type in the filter state input

### What is expected?
The state is filtered

Stack overflow error

Akryum:
- Fix #550 with `seen` Map
- Search takes into account special tokens like 'null' or '-Infinity'
- Better support for Array, Set, Map
- Maximum search depth (currently 10 levels)


src/util.js
```diff
@@ -312,44 +327,70 @@ function isPrimitive (data) {
...
     }
   }
   return match
 }
 
-function compare (mixedValue, stringValue) {
-  if (Array.isArray(mixedValue) && searchInArray(mixedValue, stringValue.toLowerCase())) {
-    return true
+function interalSearchCheck (searchTerm, key, value, seen, depth) {
+  if (depth > 10) {
```
posva: It's okay for a hotfix, but wouldn't it be better to use something like `WeakMap` to make sure we're not checking an object we have checked before?

Akryum: It's what the `seen` `Map` is for.

posva: oh lol sorry, I thought we couldn't use objects as keys in js maps, not sure why.
Why do we wait for depth to be 10, wouldn't be safe to stop if we check the same object twice while inspecting an object?

Akryum: For performance reasons, I though it would be nice to have some kind of limit.

Akryum: > I thought we couldn't use objects as keys in js maps

It's the other way around, we can use objects in `Map` but not in `WeakMap`.

posva: I've used WeakMap with objects before 😆 

About the depth, I don't get what you mean with performance, if we drop the search because we found an object we already found while looking recursively into anothe object, wouldn't that be the sortest way to stop the search?

```js
// given
{ a: { b: obj }, c: obj}
```
once we try to lookup `obj` inside of `b`, we would never look it up inside of `c`


```js
// given
var a = {}
var b = { b: { a }}
a.b = b
```
 we would only dive till `a.b.b` (or `a.b.b.a` depending on the implem)

Akryum: Okay, I don't why I had errors with `WeakMap`.
If you don't have any match, we could spend a lot of time on very large objects, I thing it's a good idea to have some kind of hard limit on recursive depth (also, the algorithm can search deeply into arrays now).

posva: Ah, okay so it was to limit deep nested objects no matter repetition.
Forget about WeakMap, Map is ok 🙂

src/util.js
```diff
@@ -313,45 +328,113 @@ function isPrimitive (data) {
...
 
+/**
+ * Searches a key or value in the object, with a maximum deepness
+ * @param {*} obj Search target
+ * @param {string} searchTerm Search string
+ * @returns {boolean} Search match
+ */
 export function searchDeepInObject (obj, searchTerm) {
-  var match = false
+  return internalSearchObject(obj, searchTerm.toLowerCase(), new Map(), 0)
```
posva: I think it would be safer to clear the map when leaving the function to avoid holding references to objects and creating a leak

PR #573, opened 2018-01-26
Escape HTML for more data

Akryum:
- Function names
- Native objects display values


Akryum:
capture d ecran du 2018-01-26 14 34 14

PR #575, opened 2018-01-27
SharedData: classifyComponent, openInEditorHost

### LINKED ISSUE #598: Open in editor not working when not served by node.
### Version
4.1.2

### Browser and OS info
All

### The problem
Hi,

This issue concerns both the vuejs devtools and the @yyx990803 [Lauch editor](https://github.com/yyx990803/launch-editor).

I'll take my example to explain the problem but it's exactly the same for every projet where vue is not served by nodejs.

I'm a Laravel user and I use laravel-mix which is just a wrapper over webpack.
Our html files are served by php, but we also have a hot reloading function which start a node server, and serve the js and css files. (see https://laravel.com/docs/master/mix)

The problem here is that the fetch here is done relatively, and so the '/__open-in-editor' route will be asked on the php server : 
https://github.com/vuejs/vue-devtools/blob/dc75c5f63d0041f7a5bed24dc7902223f0e42c2a/src/util.js#L497

So we have two options:
- If the node server is launched, and so that the '/__open-in-editor' route is up, we have to make a php route to redirect to its url (but, actually, make a php route to use vuejs devtools is really overkill).

- If the node server isn't launched and files are just built and served by nginx, we can't do anything.

Moreover, if the ide is not installed exactly where you planned, this doesn't work at all (I'm using PhpStom EAP version, which is not in the folder you supposed)

Why don't use the IDEs APIs for that ? they (almost) all have an url to show the file. 
Examples:

- "sublime"  => "subl://open?url=file://%file&line=%line",
- "textmate" => "txmt://open?url=file://%file&line=%line",
- "emacs"    => "emacs://open?url=file://%file&line=%line",
- "macvim"   => "mvim://open/?url=file://%file&line=%line",
- "phpstorm" => "phpstorm://open?file=%file&line=%line",
- "idea"     => "idea://open?file=%file&line=%line",
- "vscode"   => "vscode://file/%file:%line",

I have not the answers to all my questions, but thoughts about some (to start, fetch the absolute url from the real node server)

Thanks for your help, and your amazing job.
Matt'

### What is expected?
Component should be opened in editor.

Component is not opened, and a 404 is returned by backend.

Akryum:
New `SharedData` system that automatically synchronize data between the devtools and the backend, with persisting capabilities. Data is reactive in Vue apps.

Currently holds:
- `classifyComponents` (persisted), used by both the devtools component and the backend component highlighter
- `openInEditorHost` that can be set by the backend code (fix #598)


Akryum:
Here is the doc for `openInEditorHost`: 

---

## Customize request

You can change the request host (default `/`) with the following code in your frontend app:

```js
// App served from port 4000
// Webpack dev server on port 9000
Vue.config.devtoolsConfig = {
  openInEditorHost: 'http://localhost:9000/'
}
```

Akryum:
This may need Vue typings update.

Akryum:
@evan Changed to `window.VUE_DEVTOOLS_CONFIG`

src/shared-data.js
```diff
@@ -0,0 +1,91 @@
...
+    if (!params.hasOwnProperty('storage')) {
+      throw new Error('Missing `storage` params.')
+    }
+    storage = params.storage
+  }
+
+  // Load persisted fields
+  if (persist) {
+    persist.forEach(key => {
+      const value = storage.get(`shared-data:${key}`)
```
michalsnik: Perhaps we should use less generic storage key? Additional `vue` related prefix would rather be enough :)

Akryum: This is done on the devtools side, and thus the localStorage is already specific to the extension.

ElMassimo:
What's the timeline on this? Would be pretty useful for anyone using Webpacker to mount a Vue app in their Rails application.

Issue #442, opened 2017-10-26
Ability to copy an object to text


MikaelEdebro:
One feature that would be awesome is to be able to right click an object in the vue-devtools store panel, and select "Copy to text".
Sometimes I find it can be very useful to be able to copy for example an array of objects, and paste it to a text editor. That way I don't have to expand all the array objects one by one, and I can perform text searches of the data.

posva:
Are you aware you can copy the state already by clicking on the copy button?

MikaelEdebro:
@posva Yes I've seen this. However this copies the entire state. It would be nice to be able to copy just one single object/array/etc.

crcatala:
This would be useful to copy partial state, an for the component inspector to copy values there also. We could add to the existing inline quick edit actions in `DataField`. Maybe something like:

<img width="209" alt="copy-value" src="https://user-images.githubusercontent.com/1514479/35197549-45509c76-fea6-11e7-873a-c436efa57006.png"> <img width="245" alt="copy-value-hover" src="https://user-images.githubusercontent.com/1514479/35197550-48ac1878-fea6-11e7-813e-9fe5744b419e.png">

Would love to contribute a PR on this. Thoughts or any specific implementation you had in mind?

MikaelEdebro:
@crcatala Yes, something like that would be awesome. Don't know if anyone else than me think this would be a great feature? 
I find myself wanting to copy some object (as JSON) quite often, just to paste it in to some text editor to diff with another one.

Akryum:
I suggest adding a context menu with several items, like copy value or store as global var. Let's not clutter the state inspector too much (there are Edit buttons already).

crcatala:
@Akryum fair point, to support further actions in the future perhaps a context dropdown such as:

<img width="292" alt="more-actions-dropdown" src="https://user-images.githubusercontent.com/1514479/35351428-78735c58-0106-11e8-9ac1-9932e938b3e6.png">

PR #583, opened 2018-01-29
Add dropdown menu to DataField with Copy Value

### LINKED ISSUE #118: Right click -> Copy object
Hi, and thanks for a fantastic development tool. 

While working with objects in console, I find it very useful to right click the object, and then "Store as global variable". The object typically becomes available as temp1. Pretty much as you do when sending the vue model you´re inspecting to console as $vm. 

Some even more useful thing, is that running copy(temp1) allow me to copy the object. Copying an object and then pasting it somewhere is often so much faster than using the expand arrows etc., and typically makes me save quite a bit of time when working with data structures. 

What would be totally awesome in vue devtools, is if it was possible to simply right click on a object and then copy it directly to the clipboard. Thanks in advance for considering to make this a feature.

crcatala:
Closes #118

Adds a dropdown to Data Field with Copy Value. Can support more actions in future via this dropdown

vue-devtools-copy-value

Refs #442


crcatala:
Some items up for discussion:

* Currently only one item in dropdown menu which seems odd. Other actions can be added here in the future though is the idea.
* Leveraged the `copyToClipboard` function, but duplicated for now. Not sure how authors want to abstract this service (I see a utils dir)
* For some values such as `NaN`, `undefined`, etc the copied value is something like "__vue_devtool_nan__".
* Added an UNscoped style tag to style the popover `.popover.context-menu-popover` since it is rendered outside main app node (popper.js implementation). Not sure if we should move elsewhere.

DannyFeliz:
@Akryum when will be a new release with theses changes?

crcatala:
Thanks for getting my PR in @Akryum , appreciate all the hard work you and team are doing on the devtools! Looking forward to 4.x release 🙌 🎉

Issue #588, opened 2018-01-30
Vue panel stuck across three environments 


JustinDraude:
### Version
4.1.2

### Browser and OS info
Chrome 64.0.3282.119/ Windows 7   Mozilla 58.01/Windows 7   Electron version of Vue devtools

### Steps to reproduce
*Running Vue through npm installed package vue@2.5.13

*Laravel backend

*No Vue router, as Laravel handles routing through views

*Initial load of page (any page) works fine. Vue devtools working accordingly, inputs get updated, events tab 
can be seen, etc

*Access different page url of the website

*Panel is now non-reactive to any mouse clicks

*Components from old url page still showing

*Cannot switch to events tab or any other tab

*Cannot do anything

*Vue panel works again after closing browser tab completely and reopening

*Hard refresh doesn't fix it

### What is expected?
Vue panel to work accordingly

### What is actually happening?
Vue panel is non-responsive

---
Same issue with Mozilla plugin, as well as electron standalone. All reinstalled today, issue remains.
Three different apps were tested, making sure no existing endless loop could cause this. Same issue.
It was working fine 3 hours ago. 
Uncertain if issue is AFTER 29.01.2018 devtools update, as I don't notice when chrome updates my extensions

samydoesit:
Same here

poldixd:
Same here. VueJS Panel works and when i reload the page i can't click on any tabs like _Components_ or _Vuex_. Only if i close the Developer Console in Chrome and restart it, i can do anything again.

I tried it in Chrome and Firefox. I removed the extension and reinstall them.

fnkky:
Same hier

fnkky:
Same here, Chrome 64.0.3282.119

csicky:
When this happens there is also an error in console:

> backend.js:1 Uncaught TypeError: Cannot read property '__VUE_DEVTOOLS_UID__' of undefined
>     at backend.js:1
>     at t.a.v.on.e (backend.js:1)
>     at t.a.n.emit (backend.js:1)
>     at e.listen.e (backend.js:1)
>     at n (backend.js:1)

JustinDraude:
@csicky No error for me, perhaps project related on your end.

franzi50:
Exact same error as csicky.

ejdrian313:
same error

samydoesit:
@csicky thats the same error we get

csicky:
[...]

PR #590, opened 2018-01-30
Fix initEnv, so it doesn't crash on page refresh

michalsnik:
Redefining ownProperties on Vue.prototype causes error on page refresh, as it executes `initEnv` once again. I check for presence of one of the properties assigned there so it doesn't reinitiate it when already initialized.


Akryum:
Fix #588

PR #614, opened 2018-02-11
Make ability to send vm to console slightly more obvious

crswll:
I added a title attribute: "Available as $vm0 in the console."

Also changed `== $vm0` to `= $vm0` since it seems like something was assigned instead of just equal to something. I was also considering maybe `= window.$vm0` to make it even more clear?

Let me know what you think.


psi-4ward:
I thought about storing a reference of the focused component.
It's often hard to click through the component tree inside dev-tools but got better with "Inspect component". Why not just add a window onClick listener and store the topmost component in $vm1 ?

Voha34521:
hi

Issue #616, opened 2018-02-12
CSP unsafe-inline required to use?


lol768:
### Version
4.1.3

### Browser and OS info
Firefox  59.0b8

### Steps to reproduce
* Read README
* Attempt to use Vue dev tools on site with strict CSP headers
* Check console for errors:

`Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src”). Source: ;(function(e){let t={};if(e.hasOwnProper....`
Page:1

`Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src”). Source: ;(function(e){setTimeout(()=>{const n=do....`
Page:1

`Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src”). Source: ;(function(e){let n=null,t=0;const o={no....`
Page:1

* Re-read README.md hoping that requirement to inject scripts in this way is documented somewhere

### What is expected?
* Dev tools should operate without having to compromise on site security by allowing inline scripts
* If this isn't technically possible, this incompatibility should be documented clearly in the README

### What is actually happening?
* Dev tools cannot detect running Vue instance

Akryum:
The devtools can't communicate with the webpage without injecting a script.

lol768:
Does it have to be inline? Can you reference a `moz-extension` script, or one that's externally hosted?

chearon:
There's at least [one extension developer](https://transitory.technology/browser-extensions-and-csp-headers/) who's gotten around CSP. This was a hard error to figure out for me, and disabling CSP locally isn't really ideal either. I haven't looked at how the extension communicates with Vue yet, but there might be a way...

chearon:
Been playing around a bit with this and I've found a few ways to get the dev tools working with CSP enabled. The first two solutions require changes to `vue-devtools`:

1. In hook.js and detector.js where the `<script>` elements are created, if it did a `setAttribute('nonce', 'vue')` then for dev mode people could add `'nonce-vue'` to `script-src`. This would leave CSP on for everything else so you can still catch stray script tags before going on production
2. Firefox actually lets you evaluate stuff on the main window using `window.eval()`, so for that browser you could skip the `createElement('script')` entirely.
[...]

PR #621, opened 2018-02-16
Don't break when CSP is on in Firefox (#616)

chearon:
Turns out #616 is specific to Firefox. In Firefox you can to use `window.eval` to evaluate on the content window. This tests the browser to see if the script should be evaluated that way, and uses the old behavior for Chrome/Safari.


chearon:
Any chance someone could look at this for us lowly Firefox users? [Here](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_scripts#Using_eval()_in_content_scripts) is the section in MDN that tells you about evaluating scripts in the content window if it helps.

(Also, not sure if I did the browser detection right, not something I try to do normally but maybe unavoidable in an extension).

michalsnik:
Hey @chearon Thanks for taking the time and working on it, however I don't think this is the right solution. What I see here is omitting `unsafe-inline` CSP directive, but there is also `unsafe-eval` which usually is used alongside `unsafe-inline`. Both are strongly against all CSP reasoning.

I personally don't set those headers in dev environment, but rather on staging/production, where they actually matter.

Have you considered this approach, or am I missing something? :)

chearon:
> What I see here is omitting unsafe-inline CSP directive, but there is also unsafe-eval which usually is used alongside unsafe-inline

This still works with `unsafe-eval` on. This PR basically makes the extension follow Mozilla's advice 🙂

> I personally don't set those headers in dev environment, but rather on staging/production, where they actually matter.

That's what I did initially, but we realized that that's a dangerous route to go down because you won't catch CSP issues until after deployment.

michalsnik:
Ah, so what you're saying is that Firefox treats `eval` executed by it's extensions as something that doesn't violate `unsafe-eval` option, and it's only problem is with inline scripts injected by extensions? If yes, then your solution might well be the answer in the end :) I have to investigate it just to be sure that's the case.

chearon:
> only problem is with inline scripts injected by extensions?

Yep! I guess Chrome keeps track of where a `<script>` was created and uses _that_ to determine CSP violation, whereas Firefox prefers you to use `window.eval()`. (Side note: the [article](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_scripts#Using_eval()_in_content_scripts) implies that `window.eval()` is different than `eval()` in extensions 😐)

michalsnik:
Such a lovely quirk :D I'll try to run this solution natively on my mac tomorrow and I'll also confirm firefox detection then :) Hopefully I'll have some time in the evening.

chearon:
Any update on this? Looks like there are a few other Firefox users after all!

manniL:
:heart:

PR #625, opened 2018-02-21
(fixes #597) Allow debugging through proxy

### LINKED ISSUE #597: Use standalone version with HTTPS
Hello!
Well, the tittle says it all, is it possible to make the standalone version of the devtools use HTTPS? I tried to hack my way through it but my knowledge about Express and Socket.IO is non existent.

In our case, we use a HTTPS only dev environment and sadly this is not working.

If it were to use HTTPS by default, it'd "just work" for any HTTP and HTTPS environment.

Thanks in advance!
F

michalsnik:
This PR addresses #597 as well as some other thing by the way, like:
- port overriding - it turned out it was not entirely possible. I also updated documentation with necessary informations of how to do it.
- it makes it possible to reset default port, which might be useful while debugging using ngrok proxy, where simple link proxies to proper port on localhost and thus connection URL doesn't need additional port.

By introducing above changes it also became possible to inspect staging applications if necessary, by proxying devtools middleware through ngrok. Just as with HTTPS on localhost :)

### DIFF
--- a/shells/electron/src/devtools.js
+++ b/shells/electron/src/devtools.js
@@ -2,7 +2,7 @@ import io from 'socket.io-client'
 import { initDevTools } from 'src/devtools'
 import Bridge from 'src/bridge'
 
-const port = process.env.PORT || 8098
+const port = window.process.env.PORT || 8098
 const socket = io('http://localhost:' + port)
 const $ = document.querySelector.bind(document)
 const $intro = $('#intro')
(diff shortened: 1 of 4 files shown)


shells/electron/src/devtools.js
```diff
@@ -2,7 +2,7 @@ import io from 'socket.io-client'
 import { initDevTools } from 'src/devtools'
 import Bridge from 'src/bridge'
 
-const port = process.env.PORT || 8098
+const port = window.process.env.PORT || 8098
```
Akryum: Is this legitimate? (not a typo?)

michalsnik:
Friendly reminder @Akryum  ☝️

Issue #269, opened 2017-02-25
Root component disappears in specific situation


posva:
repro:  https://codepen.io/skyrpex/pen/mwmJRe?editors=1010 (thanks @skyrpex )

Root instance (`vm = new Vue`) created with:
- string template or render fuction
- template depending on data
- Using one component

If we modify a reactive property of the `vm`, the Root component disappears from the devtools:
<img width="380" alt="screen shot 2017-02-25 at 18 29 28" src="https://cloud.githubusercontent.com/assets/664177/23333233/5dd3194e-fb88-11e6-8268-c571a7f8bcc9.png">

Commenting that line fixes it:
<img width="447" alt="screen shot 2017-02-25 at 18 30 07" src="https://cloud.githubusercontent.com/assets/664177/23333239/79a7e7b2-fb88-11e6-87d9-dd598635adb7.png">

**Current workaround**: select the topmost Vue instance and, in theconsole, write `$vm.$root` to debug the root Vue instance

edDimensi:
have same problem.
Then i try use this computed methods, i lost <Root>
```
  computed: {
    getAlbums () {
      return this.$root.base.albums
    },
    getPosts () {
      return this.$root.base.posts
    }
  }
```

skyrpex:
I loose <Root> too, but not sure what triggers it. All I know is:
* I load my page, and there's no <root>.
* I navigate to other page (I'm using vue-router), and then everything dissapears from the vue-devtools.
* I press the vue-devtools refresh button and <root> is correctly displayed for the rest of the session.

thomthom:
I'm seeing similar issue as well when I set some properties. Haven't been able to exactly pin down the exact trigger conditions.

bobmoff:
Same issue here

MaxMilton:
Since the latest version of vue-devtools I haven't experienced this anymore. It might be worth double checking it's still an issue for you.

bobmoff:
Yes! I can see Root again, sweet!

skyrpex:
With vue-devtools 3.1.4 and Vue 2.3.4 I keep experiencing the error in my setup.

posva:
I accidentally deleted my old repro but I cannot reproduce it anymore 🙁 
@skyrpex If you manage to reproduce it in a boiled down codepen (https://codepen.io/pen?template=PWbYJO) or a simple html file, please, could you open a new issue with it?

skyrpex:
[...]

PR #637, opened 2018-03-07
Fix loosing <Root> #269

UncleBill:
We got a same issue to #269 . I don't know the real reason, but I believe this patch will fix it.

BTW, When I inspect the code, I found something, may be useful for finding the real reason @yyx990803 :

In the backend.js, when the detected `rootInstance`  is a `VueComponent`, we'll loose the `<Root>`


UncleBill:
@posva Could you have a look?

Amareis:
Can it be related to #645?

adriancast:
Do you guys have any date when is going to be updated the chrome extension?
Thanks

UncleBill:
@adriancast This patch was include in 4.1.4, which was already published to the webstore.

Issue #645, opened 2018-03-12
Duplicate Component Trees


monooso:
### Version
4.1.4

### Browser and OS info
Chrome 65.0.3325.146 / macOS 10.13.3

### Steps to reproduce
Open the `vue-devtools` extension, and navigate to the "Components" tab. You should see several identical Components trees (three, in my experience). For example:

```
<Root>
  <MyApp>
<Root>
  <MyApp>
<Root>
  <MyApp>
```

### What is expected?
`vue-devtools` should display a single Component tree, as per previous version of the extension.

### What is actually happening?
`vue-devtools` displays several duplicate Component trees.

---
This bug also appears on Chrome 64.x, and Firefox 58.x. It appears to have been introduced in `vue-devtools` version 4.1.4.

aguingand:
I have the same issue. As for me, it's happening when several of my components "`import Vue from 'vue'`"

joelbradbury:
I've got the same behaviour, introduced only in 4.1.4. The 4.1.3 devtools showed just a single component (as expected).

Interacting with any one of the component trees in the component view mirrors that action across all duplicated copies. (ie. expanding a collapsed node in copy 1, expands the same node in all other copies at the same time).

monooso:
One other thing to note. Interactions are mirrored, as described by @joelbradbury, _however_ (and it's an important however) setting a property value in one of the component trees doesn't necessarily set it in all of the component trees.

That's a real problem, and it's bitten me a couple of times already.

lusess123:
Duplicate removal rootInstances #647

mirari:
I've got the same behaviour.
![snipaste20180320_175752](https://user-images.githubusercontent.com/4995159/37647912-7d78e49c-2c68-11e8-9834-75855f65d442.png)
But my old project seems well.
I tried to copy `package.json`, `yarn.lock`, and all webpack configs, but still got the bug.
![snipaste20180320_175900](https://user-images.githubusercontent.com/4995159/37647916-7ed7be94-2c68-11e8-8257-7f1dd44b2b17.png)

ericcirone:
I'm having the same issue of multiple <Root> and components.

badpunman:
Same issues.

Also experiencing instability in some scenarios when trying to use the component name search. It just freezes the whole vue-devtools and requires a restart of the whole browser. Might be unrelated issue though, and worth separate item?

rawbin-:
[...]

PR #647, opened 2018-03-15
Duplicate removal  rootInstances

lusess123:
Just as  [React Protals](https://reactjs.org/docs/portals.html) ,a DOM node that exists outside the DOM hierarchy of the parent component may be the same rootInstance    
eg.  
      [ iview      Modal/Poptip/Dropdown/...     "transfer"    props  ](https://www.iviewui.com/components/dropdown-en#API)   
       True  to   “ Whether   to append the layer in body ”   ,      Multi   duplicate  root instances   appear  in  vue-devtools . #645 

I think the better solution is if the same root node displays the component node, and different root nodes show another root node.


Akryum:
Fixed in #711, thanks for your PR!

PR #661, opened 2018-04-13
feat: detect Nuxt.js

pi0:
Adds nuxt.js detection for chrome/firefox extensions. A lighter check for Vue.js is preferred when any signs of Nuxt detected. `window.__NUXT__` is also being checked so we can detect Vue even if for any reason app is not fully loaded yet.


quanzhiyuan:
i like that, thks

PR #714, opened 2018-07-31
Only open file if it's a full path

chrisvfritz:
Precompiled 3rd party components still need to define a `name` on components, because the `__file` property is not currently added in production. For privacy and practical reasons, we never want to add the full path to `__file` in production, but we want to start adding setting the file's basename in `vue-loader` and `rollup-plugin-vue`.

For this to work though, we'll want to start checking whether `__file` is a path here and only allow it to be opened if it is.


chrisvfritz:
@Akryum What do you think?

PR #723, opened 2018-08-01
Remember initial mutations

### LINKED ISSUE #408: Vue Devtools not capturing initial vuex commits

simplesmiler:
Fixes #408.
This is alternative to #491, that does not require any changes to Vuex.


bartlomieju:
I really like the idea it requires not changes to Vuex!

EmmanuelDemey:
This fix is already available ? Which version should we use ?

Akryum:
5.x beta

EmmanuelDemey:
Thanks

PR #782, opened 2018-09-23
feat: Add Milliseconds Display Preference and Remove Duplicate Filter…

### LINKED ISSUE #777: show event full timestamp
atm the events timestamp shows as `H:M:S` is it possible to show it all the way down to `ms` ? 

this is helpful because currently vue doesnt offer a way to check if an event was fired or not and to get around that we need to use `setTimeout` which is not the same with all the events. example https://forum.vuejs.org/t/how-to-check-if-an-event-was-fired-or-not/43828

// before
```js
export function formatTime (timestamp) {
  return (new Date(timestamp)).toString().match(/\d\d:\d\d:\d\d/)[0]
}
```

// after
```js
export function formatTime (timestamp) {
  let date = new Date(timestamp)

  return `${date.getHours()}:${date.getMinutes()}:${date.getSeconds()}:${date.getMilliseconds()}`
}
```

aprouja1:
Closes #777 and removes instances where the formatTime filter was registered more than once


src/shared-data.js
```diff
@@ -4,6 +4,7 @@ const internalSharedData = {
   classifyComponents: true,
   theme: 'auto',
   displayDensity: 'low',
+  timeFormat: 'default',
```
Akryum: I think it should also be added to the `persisted` list.

ctf0:
any updates ?

PR #800, opened 2018-10-14
Add separate devtools-client package.

tupperkion:
# Background

In experimenting with the recent addition of NativeScript support, I noticed one thing: electron was causing my builds to fail. But NativeScript apps don't even use electron, so I investigated why a desktop framework was being installed. A simple `npm list` command showed me what was happening. The `nativescript-vue-devtools` command depended on `@vue/devtools`, which depended on `electron`.

### Why is this a problem?

Having extra dependencies increases app size, and in some cases, can create issues that are hard to debug. There is no need to include electron in a mobile app.

### What does this pull request do?

I copied `shells/electron/index.js`, `shells/electron/src/hooks.js` and `shells/electron/backend.js` to a new folder: `shells/client`. I also copied the `package.json` over, and removed extra dependencies. I modified the package name to `@vue/devtools-client` to allow users to install this new backend package. I also modified `README.md` to discuss the new changes. I haven't added a README to the client shell folder yet.

### How do other people need to respond to this change?

App developers don't _need_ to do anything (see below). However, if they desire, developers currently using `@vue/devtools` as a runtime dependency (importing it using `require('@vue/devtools')`) may move `@vue/devtools` to dev dependencies, and instead import the new `@vue/devtools-client` package from the `shells/client` folder. This means that the only extra dependency being built with the app is socket.io.

### Backwards compatibility.

The `src/hooks.js` and `src/backend.js` files still exist in the electron shell, so the server behaves the same way. For backwards-compatibility with apps that use the "current" way of importing the dependency package, `@vue/devtools-client` has been declared as a dependency of the electron shell. `index.js` simply requires/exports `@vue/devtools-client`.

> If maintainers want to enforce version-likeness, feel free to revert `shells/electron/index.js` and remove the dependency.

### What's did I leave for maintainers to do

 - Add a README in `shells/client`
 - Decide whether to keep or revert `index.js` and `package.json` in `shells/electron` (see note above)
 - Whatever other changes you want to make

### DIFF
--- a/shells/client/.npmignore
+++ b/shells/client/.npmignore
@@ -1,0 +1,2 @@
+node_modules
+src
--- a/shells/client/index.js
+++ b/shells/client/index.js
@@ -1,0 +1,24 @@
+require('./build/hook.js')
+
+const target = typeof window !== 'undefined'
+  ? window
+  : typeof global !== 'undefined'
+    ? global
+    : {}
+
+module.exports = {
+  connect: function (host, port, { io, showToast, app } = {}) {
+    target.__VUE_DEVTOOLS_HOST__ = host
+    target.__VUE_DEVTOOLS_PORT__ = port
+    if (io) target.__VUE_DEVTOOLS_SOCKET__ = io
+    if (showToast) target.__VUE_DEVTOOLS_TOAST__ = showToast
+    if (app) target.__VUE_ROOT_INSTANCES__ = Array.isArray(app) ? app : [app]
+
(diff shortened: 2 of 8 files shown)


jbosse:
I like this. It also mitigates things like #804

trishaanb:
Add on please. Thank you Patricia Collier collierp118@gmail.com

Akryum:
Electron should not be imported in your app though even if you install `@vue/devtools`.

Issue #779, opened 2018-09-23
 [Typescript] add type definitions 


championswimmer:
### What problem does this feature solve?
stops typescript from complaining of absent type declarations

### What does the proposed API look like?
A very simple 

```ts
declare module '@vue/devtools' {
  const connect(host: string, port: string): void
}
```
will do. only required so that we can do 

```ts
import devtools from '@vue/devtools'
devtools.connect()
```

karanisverma:
Hey @Akryum can I pick this up?

Akryum:
Of course! :)

karanisverma:
Thanks @Akryum, Looking into this

championswimmer:
I think both host and port are optional, so my given example is not 100% correct.
So it should be `host?: string, port?: string` in the arguments

PR #836, opened 2018-12-05
[Typescript] added type definitions

karanisverma:
I have added type def for connect.
Issue: https://github.com/vuejs/vue-devtools/issues/779

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,3 +1,3 @@
 declare module '@vue/devtools' {
-    export function connect(host?: string, port?: string): void
+    export function connect(host?: string, port?: number|string): void
 }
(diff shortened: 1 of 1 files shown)


karanisverma:
@Akryum Why this PR is not getting Merged? Please let me know if there is something which need to be fixed here.

yoyoys:
I think set port to `number|string` is better.

karanisverma:
sure will make this change! Thanks @yoyoys

karanisverma:
@posva  does these changes looks good ?

posva:
I don't know. I haven't followed dev tools dev since @Akryum picked them up. I didn't know you could install DevTools as a package now. He will look at it whenever he has time. But to be honest this looks like low priority compared to buy GA and features so it may take a while

karanisverma:
Thanks @posva 🙂

championswimmer:
@Akryum 
This seems good to be merged. Very minor anyway. The types are all correct.

PR #838, opened 2018-12-05
[Feature]: RegEx Lookup in Events Tab

ericnograles:
# Summary

While looking through some StackOverflow questions, I ran into someone [who was asking how to whittle down a list of Vuex Mutations and Vue Events](https://stackoverflow.com/questions/50542948/is-it-possible-to-exclude-vue-events-and-vuex-mutations-from-vue-devtools). It appears his app can generate up to 50 events/second and it essentially crashes Vue DevTools due to the volume and on-screen rendering.

I saw that RegEx filtering was available for the Vuex tab but not the Events tab, so I took the liberty o taking a crack at applying the same approach to the Events tab.

Please lmk what adjustments I can make. Thanks!

# Screenshots

image


johnrhampton:
👍

DannyFeliz:
I need this

Issue #848, opened 2018-12-12
Not calling vuex Getters when no needed


trollderius:
### What problem does this feature solve?
A SPA website can have different store modules with specific getters for each screen, which can be associated with ajax requests. 
Executing all the getters automatically will not only execute all the associated requests, but also can create inconsistencies or very different behaviour.

### What does the proposed API look like?
Maybe getters could be collapsed and disabled by default, and call them all only on expand. Allowing to use the rest of tools normally.

Akryum:
Are you implying that you are making ajax calls in getters? :scream_cat:

trollderius:
> Are you implying that you are making ajax calls in getters?

A Handler function requests ajax once

AlexandreBonaventure:
So I have this issue with one of my app where evaluating getters on vuex:init is throwing errors and breaks everything. I am using vuex-router-sync and one the getters is written as such :
```js
export const stationId = (state, getters, rootState, rootGetters) => {
  return rootState.route.params.stationId
}
```
This getter is not meant to be evaluated on init and since vuex-router-sync add the `route` state module asynchronously it fails.
This is an easy fix :
```js
export const stationId = (state, getters, rootState, rootGetters) => {
  return rootState.route && rootState.route.params.stationId
}
```

But it feels bad to modify the codebase just for using the devtool.

PR #849, opened 2018-12-14
don't execute getters on vuex init

cjpearson:
Relating to issue #848 this change avoids cloning the getters when creating the initialStore. As a result the getters will no longer all be executed on init.


Akryum:
Executing the getters is necessary to save their values alongside the initial state of the store.

cjpearson:
Is there a reason we need to save the values instead of recomputing them? Executing them immediately can cause issues with getters that should not be run on the initial state.

Issue #757, opened 2018-09-05
Error thrown when editing Vuex state via devtools in strict mode


mserajnik:
### Version
5.0.0-beta.1

### Browser and OS info
Chrome 69 / macOS 10.13.6

### Steps to reproduce
1. Enable strict mode in Vuex
2. Edit the state via devtools

### What is expected?
The state should be changed without throwing an error (since we are manually and knowingly changing the value via devtools).

### What is actually happening?
The state is changed, but Vuex throws an error:

```
Error: [vuex] Do not mutate vuex store state outside mutation handlers.
```

---
It's really only a minor annoyance – not sure if it this can be changed at all (or if it's worth the trouble).

PR #829, opened 2018-11-18
#773: Fix payload on replaying mutation

Sp1ker:
closes #773 , closes #792, closes #802 
All handlers are `wrappedMutationHandler ` with only 1 argument - `payload`

PR #866, opened 2019-01-22
Fix mutations replay

### LINKED ISSUE #792: Time travel passes state instead of payload
### Version
5.0.0-beta.3

### Browser and OS info
Chrome 69 / OSX 10.13.6

### Steps to reproduce
 - open https://new-issue.vuejs.org/?repo=vuejs/vue-devtools
 - click the button twice
 - time travel to first mutation
 - time travel to second mutation
 - check console

### What is expected?
State changes as per the mutation code

Mutation is called with state as payload

nialna:
Not sure if this fix makes sense, as I don't know the rest of the code, but creating a PR just in case.

Fixes #792. Also related to #757: The replay mutations makes Vue throw an error because it's not allowed to edit the store while not commiting. I added a line just before and after the mutations call to set `store._committing` to true, so it allows to commit.

### DIFF
--- a/src/backend/vuex.js
+++ b/src/backend/vuex.js
@@ -146,7 +146,8 @@ export function initVuexBackend (hook, bridge) {
       // Replay mutations
       for (let i = snapshot.index + 1; i <= index; i++) {
         const mutation = mutations[i]
-        mutation.handlers.forEach(handler => handler(state, mutation.payload))
+        store._committing = true;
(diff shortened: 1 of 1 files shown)


Sp1ker:
#829

Akryum:
Fixed in https://github.com/vuejs/vue-devtools/commit/9305a517fd3d9ed19f485396c9e31b1a8425abb1

PR #870, opened 2019-01-26
Show property name and tag for transition element (and other elements)

1099511627776:
Show property name and tag for transition element (and other elements)


Akryum:
Those attributes are already displayed on the props in the component inspector and I don't think it's a good idea to clutter the component tree with more attribute than just the `key` (which is a special attribute that doesn't appear anywhere else).

1099511627776:
The main idea is to speedup the search of component. If one has many transitions in tree then he has to open all of the transition elements to find one element.

PR #895, opened 2019-03-04
Add instructions for using puma-dev to proxy devtools

brendon:


felixbuenemann:
I'd be very careful in advocating a tool that installs a local root CA.

While these tools are well intended and not harmful in themselves, they open your system up to MITM attacks of arbitrary https connections.

I think you're much better off in installing a single self-signed certificate for localhost and trusting it than having a private root CA with a user-accessible private key.

brendon:
Thanks @felixbuenemann. I'm afraid I can't figure out the exact nature of the potential attack but I'm sure you're right. Could you give an example of how this could be exploited in real life? I'm keen to know just out of interest :)

felixbuenemann:
If you have a local CA that you trust, this CA can be used to generate valid certificates for any website.

Depending on how the CA cert was generated, it might also be able to sign software, so it can automatically open firewall ports and avoid warning prompts on a Mac.

So if for some reason your CA private key falls into the wrong hands (eg. a hacked npm package), it can be used to trick your computer into running software, redirecting websites or hijacking https traffic.

The risk of using a local root CA is fine for a security researcher that legitimately needs to inspect TLS connections, but it's not something I would advocate to developers that might not be deeply familiar with how PKI works and what risks are involved by trusting a self-signed CA.

brendon:
Thanks @felixbuenemann :) That makes sense to me. I might hit up the puma-dev team about your thoughts to see what they think. It might change their mind about that aspect of their software since the SSL config comes with a default install.

Feel free to close this PR if you'd rather not add the documentation :)

PR #906, opened 2019-03-17
fix: CSP inline style violations (fix #905)

### LINKED ISSUE #905: Inline Styles cause Content Security Policy Violation
### Version
5.0.0-beta.4

### Browser and OS info
Chrome 72 / macOS 10.14

### Steps to reproduce
When the vue-devtools inspector is used on a website that has Content Security Policy deployed and the policy does not whiltelist inline styles using the insecure 'unsafe-inline' policy, the following error is logged in the console over and over:

> Refused to apply inline style because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-BLZdMv2gh2vNZKxp7hehCQqh6sIu14k4XWIywP6 mJA='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.

Note that the hint to whitelist a sha256 is misleading, since it won't work. Only inline `<style>` tags can use hashes, inline style attributes are not supported.

See https://bugs.chromium.org/p/chromium/issues/detail?id=546106 for details on the misleading error message.

### What is expected?
Vue-devtools should work fine with CSP enabled.

Vue-devtools inspector styles are not applied due to the unnecessary usage of inline styles.

---
This problem can easily be fixed by converting the affected code to use the JavaScript DOM API to create DOM elements and set the styles using the style property of the element.

The current problem only exists because the extension uses html snippets with inline styles that are inserted as `innerHTML`.

I will work on a PR to fix this problem.

felixbuenemann:
Content Security Policy prevents the usage of inline styles on DOM elements, unless the insecure 'unsafe-inline' rule is used. This can be worked around by setting styles through the JS DOM API.

Fixes #905

### DIFF
--- a/src/backend/highlighter.js
+++ b/src/backend/highlighter.js
@@ -44,10 +44,19 @@ export function highlight (instance) {
 
   init()
   if (rect) {
-    let content = ''
+    const content = []
     let name = instance.fnContext ? getComponentName(instance.fnOptions) : getInstanceName(instance)
     if (SharedData.classifyComponents) name = classify(name)
-    if (name) content = `<span style="opacity: .6;">&lt;</span>${name}<span style="opacity: .6;">&gt;</span>`
+    if (name) {
+      const pre = document.createElement('span')
+      pre.style.opacity = '0.6'
+      pre.innerText = '<'
+      const text = document.createTextNode(name)
+      const post = document.createElement('span')
+      post.style.opacity = '0.6'
+      post.innerText = '>'
+      content.push(pre, text, post)
+    }
     showOverlay(rect, content)
   }
 }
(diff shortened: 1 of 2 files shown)


felixbuenemann:
The tests are green and I have manually tested the vue inspector.

However  I have not yet manually tested the changes to the vue toast, since I don't know how to trigger it, so it would be great if someone could test it or provide instructions on how to trigger the toast.

felixbuenemann:
Also note that this changes the signature of the highlighter.js:showOverlay function to accept an array of dom elements instead of a raw html string.

I didn't see any other consumers of this function, so it should be fine.

Issue #972, opened 2019-04-15
Style issue on Settings Tab


wormwlrm:
### Version
5.1.0

### Browser and OS info
Chrome 73.0.3683.103(64bit)/ Mac Yosemiti 

### Steps to reproduce
1. Open `vue-devtool`
2. Go to `Settings` tab
3. Even there is content to scroll, I cannot scroll down.

### What is expected?
Settings tab should be scrollable

### What is actually happening?
Settings tab is not able to scroll

PR #973, opened 2019-04-15
fix: style issue on Settings Tab

wormwlrm:
I created issue #972 before.

Settings tab is not able to scroll down.

스크린샷 2019-04-15 오전 9 57 49

Because even there is a CSS to make this tab as scrollable, 
cascading rule makes `overflow: auto` ignored.

So I checked if there is a same class name with `settings` and put `!important` to make high priority.


Akryum:
Thanks for your PR! :+1:

PR #991, opened 2019-05-07
Fix required prop warning in settings tab

simsim0709:
This pr fixes required prop warning in "settings" tab

Screen Shot 2019-05-07 at 4 04 13 PM


Akryum:
Thanks for your PR! However, this seems to be an issue in vuejs/ui instead. I'll take a look at it.

PR #1036, opened 2019-07-16
Feature/clear all messages vuex events

mikaelhadler:


Akryum:
What's is the difference with the `Commit All` button?

mikaelhadler:
@Akryum `` `` ClearAll`` has not attempted to alter any state in history, just clean messages in history. The `` `Commit All``` looks like me to update information in store or something like that. The current idea for `` `Commit All``` is the same think / action to clear history? It's not clear to me.

mikaelhadler:
@Akryum Hey dude, any considerations about that?

Akryum:
I still don't see the difference with `Commit All`...

mikaelhadler:
So, that's okay : ], i'm close this PR, thanks for the reply

PR #1045, opened 2019-08-06
fix rawTypeRE regexp

huangbuchao:


huangbuchao:
this has a careless mistake. ^_^

Akryum:
Thank you for your PR! However, there is no issue with the code you modified. The `]` char is taken literally since the first `[` is escaped. There is no need to escape `]`.

Issue #1027, opened 2019-06-28
Visual bug on routes list when `redirect` is a function


sharkykh:
### Version
5.1.1

### Browser and OS info
Chrome 75.0.3770.100 x64 / Windows 10.0.17763 Pro x64

### Steps to reproduce
Have a route with `redirect` as a function, for example:
```js
{
    name: 'editShow-legacy',
    path: '/home/editShow',
    redirect: to => ({
        name: to.name.replace('-legacy', '')
    })
}
```

### What is expected?
I'm assuming the same thing as on the meta (right) panel.
Or maybe something a little shorter like: `ƒ redirect(...)`

### What is actually happening?
![chrome_cWLNH7WHUC](https://user-images.githubusercontent.com/10238474/60340070-478b0680-99b3-11e9-9c7b-35661c1103eb.png)

PR #1086, opened 2019-10-18
Fixes: #1027 fixed visual bug on routes list by adding a computed pro…

rudy750:
…perty that checks for a custom display

Used a computed property to check if the _custom.display option exists. In that case return that to the template and use v-html to properly render the text.

image


Akryum:
Hi! Thank you so much for your PR! Another solution has been implemented in https://github.com/vuejs/vue-devtools/commit/f7e2b099c973e2a5ad394c4ecd2f3da47710c1ee

rudy750:
That is definitely fancier! Happy to be involved will look for another open bug to fix

PR #1128, opened 2019-12-12
fix: doesn't transpile after monorepo, break on IE

miaulightouch:
image

After monorepo, all package would store in `node_modules` and `buble` wouldn't touch them. if just include all `@vue-devtools` packages for transpiling, then it would be failed too by the es6 syntax that `buble` unsupported like async/await. 

so I introduce the babel-loader for transpiling pure js, and it looks fine, though I don't know it wouldn't be too bloated for this project.


maxpaj:
Is it possible to use an older version until this is merged and published?

miaulightouch:
5.1.1 may be the latest version that work on IE

Akryum:
Tests are not passing.

kspackman:
What's the status of this? We'd love to get this working in IE again.

PR #1138, opened 2020-01-01
Fix ActionHeader icons display for small width

AgentX001:
Before:
image
After:
image


mehanik-pyx:
What can we do to have this pull request reviewed and merged?

Issue #607, opened 2018-02-05
Component name is outside of view for large components


ffxsam:
### Version
4.1.3

### Browser and OS info
Chrome 64 / macOS 10.13.3

### Steps to reproduce
1. Bring up a page with a large component that is only halfway visible
2. Open Vue DevTools, click "Select" in the Components tab
3. Mouse over the large component

### What is expected?
I expect the component name to show up

### What is actually happening?
The component name is not visible because it's centered within the component, so it's off-screen.

---
This should probably check the browser window constraints and reposition the component name accordingly.

PR #1161, opened 2020-02-06
make select text fit in the window

e-krebs:
Addressing issue #607.

When a component is larger than the screen,
When using the select tool,
The name can appear outside of the visible screen (since it is centered in the component).

_**example**_:
The simplest example is setting `style="width:300vw"` on a component.

| Before | After |
|--------|-------|
| image | image |


e-krebs:
tests are passing locally 🤷‍♂️

PR #1243, opened 2020-07-24
feat(next): add store as global

pikax:
image


Akryum:
I'm not sure it will work as is, since the window of the devtools is the one in the browser devtools panel, not the one in the actual page.

PR #1258, opened 2020-08-22
feat(vue2): Add meta field into routes panel

### LINKED ISSUE #1031: Routing > See/Change Meta and Props values

We have stuff like this in our `@/routes/index.js`:

```js
{
  path: '/labs/:id',
  name: 'labDetails',
  component: LabDetails,
  props: (route) => ({
    id: route.params.id
  }),
  meta: {
    permission: 'readWriteDeleteLabs',
    showModal: false
  },
  children: [
    {
      path: 'reservation',
      name: 'labReservations',
      meta: {
        permission: 'readWriteDeleteLabs',
        showModal: false
      }
    },
    {
      path: 'reservation/:reservationId',
      name: 'labReservation',
      components: {
        modalBody: LabReservationDetails
      },
      meta: {
        permission: 'readWriteDeleteLabReservations',
        showModal: true
      }
    }
  ]
}
```

But the Vue-DevTools only shows `path` and `name`. It would be useful to be able to see the values for props and meta. It would be even better if you could change these values. We have our modal's visibility tied to
```js
computed: {
  showModal: function () {
    return this.$route.meta.showModal;
  }
}
```
So being able to toggle the value in Meta lets me change the state of the component.

.

freedomlang:
Fix #1031

### DIFF
--- a/packages/app-frontend/src/views/routes/RoutesMeta.vue
+++ b/packages/app-frontend/src/views/routes/RoutesMeta.vue
@@ -59,6 +59,9 @@ export default {
       if (routeData.name && routeData.name !== UNDEFINED) {
         data.name = routeData.name
       }
+      if (routeData.meta && !this.isEmptyObject(routeData.meta)) {
+        data.meta = routeData.meta
+      }
       if (routeData.component) {
         const component = {}
         // if (routeData.component.__file) {


DannyFeliz:
I would love to see this merged

J141:
Same. The code seems all ready to go, and displaying meta would make debugging routes easier.

TheJaredWilcurt (approved):

barayuda:
Hi folks, this feature have merged? cc @TheJaredWilcurt

PR #1265, opened 2020-09-02
feat(timeline): allow logType for errors and warnings

posva:
Screen Shot 2020-09-02 at 15 22 53


Akryum:
Did you try with the icons?

posva:
Yeah, the circle is too small for the icon. Maybe a different one would work. We could also draw something different from a circle and differentiate by shape on top of color:

Akryum:
Sounds like a good first implementation. Maybe you can delete the icon images then.

posva:
Should be good as long as the PR is squashed

packages/app-frontend/src/features/timeline/TimelineView.vue
```diff
@@ -125,11 +127,13 @@ export default {
       updateEventPosition(event, g)
       g.y = 16
       event.g = g
-      if (selectedEvent.value === event) {
-        drawSelectedEvent(event)
-      } else {
-        drawUnselectedEvent(event)
-      }
+      setTimeout(() => {
```
Akryum, line 130: Why the timeout?

posva: ah yeah, I forgot about this one and didn't put a comment! I realized that stackedEvents do not appear immediately, so some stacked events would render the first event meaning that one normal and an error would render as normal. I imagine the problem is somewhere else, where the call to `addEvent` is done but I didn't dig more 😅

PR #1269, opened 2020-09-03
feat(component inspector): allow tags for component inspector

kazupon:
component-tag-info

`walkComponentTree` hook allows you to add tags to the `payload.componentTreeData`.
For the i18n use case, we can check which components have i18n resources associated with them.


kazupon:
We might refactor 'CustomInspectorNodeTag' naming. (e.g. InspectorNodeTag)

Akryum (approved):

posva:
I tried out by doing a very naive add of tags:

```js
      console.log('walkComponentTree')
      api.on.walkComponentTree(payload => {
        console.log('walkComponentTree', payload.componentTreeData.isRouterView)
        payload.componentTreeData.tags.push({
          textColor: 0xffffff,
          backgroundColor: 0,
          label: 'router-view',
        })
      })
```

and I thought it could end up adding the same tag multiple times if the user doesn't check the existence of the tag because of how `walkComponentTree` is called. Is this okay?

ScreenFlow

kazupon:
Hmm, :thinking:
I didn't take into account that `walkComponentTree` will be called more than once.
I'll have to think about the implementation logic for this feature. I'll give it some more thought. :)

kazupon:
I've added `AppRecord` to `Contenxt`.

@posva 
`AppRecord` have `instanceMap`, and It has a Component instance of the form `app_id:component_uid` as its key (e.g. `1:2`).
`walkComponentTree` payload has the `componentTreeData`, and it has `id` like  the `app_id:component_uid` format.

The following code (vue-i18n case) sets `tags` to keep idempotency, as shown in the following example, there should be no problem even if more than one `walkComponentTree` is called:

```ts
      function updateComponentTreeDataTags(appRecord: AppRecord, treeData: ComponentTreeNode): void {
        const tag = {
          label: 'i18n',
          textColor: 0x000000,
          backgroundColor: 0xFFCD19
        }
        let insntance = appRecord.instanceMap.get(treeData.id)
        if (insntance && insntance.vnode.el.__INTLIFY__) {
          treeData.tags = [tag]
        }
        for (const node of treeData.children) {
          updateComponentTreeDataTags(appRecord, node)
        }
      }

      api.on.walkComponentTree((payload, ctx) => {
        updateComponentTreeDataTags(ctx.currentAppRecord, payload.componentTreeData)
      })
```

@Akryum 
What do you think?

kazupon:
The above code works for the vue-i18n use case.
vue-devtools-tags

posva:
Is this the whole PR? I cannot find where `currrentAppRecord` is set. Was it already set but not typed?

kazupon:
>Is this the whole PR? 

Yes.
I've added it to `@api/devtools-api` in this PR.
https://github.com/vuejs/vue-devtools/pull/1269/files#diff-062ed9deb4a8078fea9bfbe2a5c6d8e2R5

You can downcast `ctx` type to `any` and you can use it. 😎
(Actually `ctx` can be used because devtools backend passes it to the `currrentAppRecord`)

posva:
So was `currentAppRecord` already set but not typed?

PR #1270, opened 2020-09-05
docs: add architecture diagrams and plugin docs skeleton

kazupon:
vue-devtools-architecture


bencodezen (approved):
Thanks so much for this @kazupon! It looks awesome!

PR #1282, opened 2020-09-18
build(api): add bundling to support tree shaking

posva:
This is quite a lot but the important part is to make `@vue/devtools-api` tree shakable by exporting an ESM version of the library as well as changing some of the enums. But I thought exposing only one version was not really viable, that's why I added rollup to create multiple versions. We can also add common js version.

- I also changed some of the enums into const to enable them to be tree shaken https://github.com/microsoft/TypeScript/issues/27604
- Wrapped some global variables into functions

This probably needs some discussion. I kind fell like adding rollup is overkill but maybe it's necessary to be able to use it in different scenarios (bundlers, browsers)


posva:
I simplified the PR by removing rollup. These changes effectively make the devtools fully tree shakable. I tested it on the router on the branch feat/devtools

PR #1297, opened 2020-10-12
Fix: action icon size

yzen-dev:
When the width of the vue-devtools is not large, the action icons disappear
Example of an error is shown in the screenshot.

image


software-provider:
1

PR #1397, opened 2021-04-05
fix(types): Add Missing Methods & Props to "./packages/shell-electron/types"

joshmossas:
Hello,

I noticed that the `init` method exported out of `packages\shell-electron\index.js` was missing from the type file. This pull request fixes that. 

Additionally the type information for the `connect` method was missing the `{ io, showToast, app}` object as an optional prop. This pull request adds it as a parameter named "options", where the option properties are defined in the `ConnectOptions` interface.

You'll also notice that this pull request adds `vue` as a dev dependency so that the `Vue` type can included in the type definitions

## Before
Before this change Typescript show type errors

image
image

## After
Typescript errors are gone
image


liangmeisheng:
nice

PR #1428, opened 2021-04-29
Fix callHandlers is not a function

### LINKED ISSUE #1422: TypeError: pluginOn[k].callHandlers is not a function
### Version
6.0.0-beta.8

### Browser and OS info
Firefox & OSX

### Steps to reproduce
Tried to enable it on a older Vue 2 project but not able to detect app, but works fine for newly created boilerplate from Vue CLI  

### What is expected?
Should detect Vue 2 app

Not able to find Vue 2 app

image

cgarnier:
#1422
Usage of `in` iterate on custom prototype methods. `of` should fix it.
close #1422


cgarnier:
Not enough

PR #1490, opened 2021-06-21
fix: detect nuxt3

danielroe:
This PR adds hotfix support for devtools with Nuxt3. Note, devtools will be enabled in production for Nuxt3, just as it seems it is with Vue3, see [these lines](https://github.com/vuejs/vue-devtools/blob/next/packages/shell-chrome/src/detector.js#L31-L41).

This change is necessary for Nuxt3 because:
* `$root.constructor` does not get `Vue` any more in Vue3
* even if we polyfill Vue at `$root.constructor`, `Vue.config.devtools` is not (yet) valid for Vue3

### DIFF
--- a/packages/shell-chrome/src/detector.js
+++ b/packages/shell-chrome/src/detector.js
@@ -16,7 +16,7 @@ function detect (win) {
       let Vue
 
       if (window.$nuxt) {
-        Vue = window.$nuxt.$root.constructor
+        Vue = window.$nuxt.$root && window.$nuxt.$root.constructor
       }
 
       win.postMessage({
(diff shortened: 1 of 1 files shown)


Akryum:
>  Note, devtools will be enabled in production for Nuxt3, just as it seems it is with Vue3

The devtools are not enabled in production by default: https://github.com/vuejs/vue-next/tree/master/packages/vue#bundler-build-feature-flags

Akryum:
I fixed devtools detection in https://github.com/vuejs/devtools/commit/9d85d2d8fd200481498c6e9190e45b6a0db519b8

Nuxt 3 will need to update to Vue 3.2.14+

Akryum:
cc @Atinux

PR #1499, opened 2021-07-01
docs: Plugins Guide grammar

king-11:
grammar correction in sentence

### DIFF
--- a/docs/plugin/plugins-guide.md
+++ b/docs/plugin/plugins-guide.md
@@ -10,7 +10,7 @@ The API is only available in Vue Devtools 6+
 
 ## Architecture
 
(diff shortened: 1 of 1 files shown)


Akryum:
Thanks! Isn't technically "devtools" plural?

king-11:
If we think of the extension panel it's a single entity so  using `is` seemed more apt to me considering we are talking about a single entity that user is interacting with. But I can make the change if you feel we should use `is` 😅

Akryum:
I've been using `are` in the rest of the docs so I guess we should stick to it or change it everywhere 😅

king-11:
Made the changes for `are`

Issue #1518, opened 2021-07-19
api.on.visitComponentTree not work in vue2.x project


ISFhr:
I fond that the handler of visitComponentTree api did not emit when devtools is init

I use it by this way:

> devtool.ts
![image](https://user-images.githubusercontent.com/30992121/126132411-db4ddf87-d280-4ceb-a35a-1c5ee001a067.png)

> setup.ts

![image](https://user-images.githubusercontent.com/30992121/126132300-4fca03df-07db-412e-8511-34f5aed14758.png)

And in index.esm.js, I add 'console' in visitComponentTree-api's handler, it didn't work, when I tried another api , called 'editComponentState', it finally consoled '2222'
> index.esm.js
![image](https://user-images.githubusercontent.com/30992121/126132538-33b213d4-c7a0-45e6-afbd-38ffe6ba336f.png)

I read the code source and fount that , in '@vue-devtools/app-backend-api/src/hooks', when add another plugin , register an events it seems has some permission check encase outside the handler? How can I sold out this problem? I’ll realy appreciate if smoebody can help me ~

![image](https://user-images.githubusercontent.com/30992121/126133935-c6479342-b0b2-41e2-b01a-23c7696ca4cb.png)

ISFhr:
I know when the 'disableAppScope' is false, devtool will check the consistent of vue instance in vue project and devtool's context. So in vue2.x project, I did not use the 'Vue.use', but inject the vueInstance in plugin's install method, just like this:
 
![image](https://user-images.githubusercontent.com/30992121/126135246-9630974f-3428-4dc8-95f9-2c6e86c71b6f.png)

then I keep the vue instance tobe equal, but the rule of  'hasPluginPermission' I still didn't found some way to debugging

ISFhr:
And I also need another question，how can I debugging the devtool in development a new devtool-plugin ?

ISFhr:
oh, I found that in "packages/app-backend-vue2/src/components/tree.ts" didn't trigger the "visitComponentTree" hook, but in "app-backend-vue2" it has:

> packages/app-backend-vue3/src/components/tree.ts
![image](https://user-images.githubusercontent.com/30992121/126319373-6cb93d07-c10a-4e90-b235-92495cc2662b.png)

PR #1524, opened 2021-08-01
fix: Call `api.visitComponentTree` in Vue2's backend api

a20185:
Referred in #1518

In **Vue 2.x** projects, backendApi `api.visitComponentTree` is not called in component tree 's  `capture` method

If it was not intended to drop this function call by design, then it was a mistake since it never triggers `hook.on.visitComponentTree` when traversing through the component tree,  leading the userland plugin failed to perform component parsing.

It's deeply needed in our team because we are developing a custom plugin strongly relied to it.
Please consider if it's possible to fix this feature in the `app-backend-vue2` package, thanks a lot!


a20185:
@Akryum Please consider if the bugfix is possible, thanks!

PR #1536, opened 2021-08-14
fix(vue2): fix async capture errors

a20185:
Hi, @Akryum.
I've found that the `async capture` method (packages/app-backend-vue2/src/components/tree.ts)  introduced  last week lead to some cases on Vue 2. 

The Vue2 version of devtools was unable to open the **App** node, report the component structure is illegal:

截屏2021-08-14 下午6 16 26

When headed towards the devtool's console, it triggers a warning of TypeError of promise not been correctly unwrapped.

截屏2021-08-14 下午6 16 43

I found that the error is probably caused by the asynchronous calls inside the   `capture` method not been properly handled in order. Hence I've  composed the function calls with `await` keyword, which solved the problem

截屏2021-08-14 下午6 11 19

### DIFF
--- a/packages/app-backend-vue2/src/components/tree.ts
+++ b/packages/app-backend-vue2/src/components/tree.ts
@@ -175,7 +175,7 @@ async function capture (instance, index?: number, list?: any[]): Promise<Compone
       : instance.componentInstance ? [capture(instance.componentInstance)] : [])
 
     // await all childrenCapture to-be resolved
-    const children = (await Promise.all(childrenPromise)).filter(Boolean)
+    const children = (await Promise.all(childrenPromise)).filter(Boolean) as ComponentTreeNode[]
 
     const treeNode = {
       uid: functionalId,
@@ -194,7 +194,7 @@ async function capture (instance, index?: number, list?: any[]): Promise<Compone
       inactive: false,
       isFragment: false // TODO: Check what is it for.
     }
-    return await api.visitComponentTree(
+    return api.visitComponentTree(
       instance,
       treeNode,
       filter,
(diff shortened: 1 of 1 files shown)


Akryum:
That was wip, thanks for continuing the conversion from Vue 2 tree scanning to async 😻

packages/app-backend-vue2/src/components/tree.ts
```diff
@@ -191,7 +194,7 @@ async function capture (instance, index?: number, list?: any[]): Promise<Compone
       inactive: false,
       isFragment: false // TODO: Check what is it for.
     }
-    return api.visitComponentTree(
+    return await api.visitComponentTree(
```
Akryum, line 197: Await in front of return is unnecessary

packages/app-backend-vue2/src/components/tree.ts
```diff
@@ -270,7 +273,7 @@ async function capture (instance, index?: number, list?: any[]): Promise<Compone
       backgroundColor: 0xff8344
     })
   }
-  return api.visitComponentTree(
+  return await api.visitComponentTree(
```
Akryum, line 276: Await in front of return is unnecessary

Akryum (changes requested):

a20185:
My pleasure. Thanks for providing devtool's openapi. It was a fantastic ability for us! 😄

packages/app-backend-vue2/src/components/tree.ts
a20185: Resolved in 93f94fda5dc5a780c53fe57112f2bcdb9b1360f6

packages/app-backend-vue2/src/components/tree.ts
a20185: Resolved in 93f94fda5dc5a780c53fe57112f2bcdb9b1360f6

Issue #1538, opened 2021-08-14
can't import effect


zhangenming:
### Version
6.0.0-beta.15

### Browser and OS info
chrome 92

### Steps to reproduce
```js
<template>
  <div id="app">none</div>
</template>


<script setup>
import { effect } from 'vue';
</script>

```
![image](https://user-images.githubusercontent.com/21235555/129449426-cf6e3e43-55f7-4dcc-915e-b90b8e46b2c4.png)

### What is expected?
work

### What is actually happening?
not work

konpeki622:
It seems like the attribute `raw` of `ReactiveEffect` has been removed by PR [#4001](https://github.com/vuejs/vue-next/pull/4001) , we should update that in devtools.

PR #1542, opened 2021-08-18
fix(dataField): 'effect.raw' has been unavailable

konpeki622:
fix the issue #1538


konpeki622:
Thank you for your review @Akryum . I found that `effec.raw` still be used instead of `effect.fn` in previous version of vue3. Should I make it compatible?

Issue #1367, opened 2021-02-23
Remember the last expanded component structure


chesterlaykin:
### What problem does this feature solve?
I use vue devtools a lot, but I alwas repeat the same thing:  I need to select and expand a lot of nested properties, in the same way.
So I choose a component and inside "data", I for example scroll down to "form", expand it, I look for some property and expand it. It amounts to a lot of repetitive work :)

When the structure is expanded, I would like this expanded view to be stored, and the next time it is opened automatically. So it "remembers" the last expanded structure.

And if the current page doesn't match the stored structure, it will go into default state.

Maybe it could be smart and store a few of such expanded views for different url:s, that is matched to the current url. So you could switch between two different pages that has different structures.

### What does the proposed API look like?
In "Settings" there could be a "remember the last expanded component structure" (or something).

Akryum:
Work is going on on the rewritten devtools to save and restore the view between page refreshes as much as possible.

chesterlaykin:
Aha, nice!

PR #1556, opened 2021-09-18
fix(vue3): complete component tree after refreshing

konpeki622:
When I select a component with deep nesting and refresh the page, some parts of the component tree is lost.
before:
refresh1
after refreshing:
refresh2

So I try to fix it and this might be related to #1367 .

### DIFF
--- a/packages/app-frontend/src/features/components/ComponentsInspector.vue
+++ b/packages/app-frontend/src/features/components/ComponentsInspector.vue
@@ -59,7 +59,7 @@ export default defineComponent({
     // Refresh
 
     function refresh () {
-      requestComponentTree(null)
+      requestComponentTree(selectedComponentId.value)
       loadComponent(selectedComponentId.value)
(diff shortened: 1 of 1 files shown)


packages/app-frontend/src/features/components/ComponentsInspector.vue
```diff
@@ -59,7 +59,7 @@ export default defineComponent({
     // Refresh
 
     function refresh () {
```
Akryum, line 61: This function is called when clicking on `Force refresh` button in the Vue devtools. Not sure this is related to the issue you describe. Could you open a new GitHub issue with a reproduction so we can discuss it and try fixing it?

konpeki622: All right. I will reconfirm this code. if it can be reproduced I will create a issue . Thank you for your review.

PR #1573, opened 2021-09-28
spread operator compatible

lileiseven:


Akryum:
I'm not sure why this would be necessary? I don't want to go back to older JS, if your plugin users target older browsers they should transpile their code with Babel or similar.

Issue #1555, opened 2021-09-16
Capability of copying the component name


Daniel4Digital:
After clicking on a component, on the right side we can see the props, data, computed, and on that side on the top we've the Components Name and a field to filter inspected data.

If we click on components name nothing happen at the moment, it could copy the component name. Would be very handy.

best regards

jacekkarczmarczyk:
Copying should be also possible from the bottom panel, there's a possibility to select the component name, but Ctrl-C doesn't work

![image](https://user-images.githubusercontent.com/15625235/152056723-646e449c-a0cc-44f0-91d9-0fb4f581c64f.png)

Akryum:
@jacekkarczmarczyk I just tried copying the component with success. Could you try again?

jacekkarczmarczyk:
it doesn't work for me, 6.0.12, tried to copy some content from the website to make sure ctrl-c works
i can copy component name from the menu, but not with ctrl-c combination

Issue #1351, opened 2021-01-24
Configuration to optionally hide `AsyncComponentWrapper`


buzzysin:
### What problem does this feature solve?
I have a small app (about 5 or so components) I am testing that has a few too many `AsyncComponentWrapper`s to unfold. Hopefully this option will reduce the amount of times that the user has to see redundant components in the tree - and if it *is* necessary for the user, then they can keep it on.

### What does the proposed API look like?
Perhaps a checkbox in the Settings menu labelled "Hide AsyncComponentWrappers from tree".

- [x] Hide `AsyncComponentWrappers` from tree - this option will hide components that are wrapped by `defineAsyncComponent` wrappers, showing only the children.

I have no idea about the implementation of the devtools so I'm unable to provide any code samples.

bittermeatball:
Hi @Akryum, can you help me to figure out how to filter out functional components (JSX) in vue devtools?

I've already asked this on Stackoverflow
https://stackoverflow.com/questions/73939172/how-to-hide-functional-components-jsx-in-vue-devtools

> I'm currently using JSX for Vue component, but currently, the devtool would show **a very ugly nested hashed functional component tree** like this, **is there any way to avoid this**? As far as I know, Vue devtools doesn't provide us an option to hide these functional components, therefore this would be very annoying
> 
> I'm building a UI package for Form module, I want to use JSX for Vue because that would make contribution much easier when a lot of people from React.js can jump in and help. My package: https://www.npmjs.com/package/formkl

![image](https://user-images.githubusercontent.com/46400321/193647855-114e1d91-26de-4b14-93d1-262a316bf8dd.png)

lefreet:
anything new here ?

PR #1588, opened 2021-10-20
feat: UI to copy component names

konpeki622:
According to #1555 , I added UI button so that it can copy the component name, which is referred to the state data operation like this:

Thank you for your code review. The style could be optimized. And also I noticed the issue #1351  mentioned that we would like to add UI to filter out components. That can be added below the button list I think.


Akryum:
Thank you for your PR! However, I implemented a different solution here: 5d84a8b96f1f7763e0fa41ad4bd42a33d75f3d60

PR #1593, opened 2021-10-29
ci: :construction_worker: add zip files to github releases

### LINKED ISSUE #1574: Add prebuilt extension archives to release artifacts
Those who cannot install from the browser extensions stores, and might also not be able to build it themselves, it would make it trivial to just download the package and manually install extensions. It might also be useful for diagnosing issues, to have someone be able to easily manually install it directly from a release.

See the darkreader release github workflow as an example: https://github.com/darkreader/darkreader/blob/master/.github/workflows/tagged-release.yml

josh-hemphill:
closes #1574 

adds a release github workflow that runs through the build and test process on releases but attaches the zip files as artifacts to the github release, so people in offline environments have an easy way install it with a single file.

Darkreader already does this; I've used their github workflow as a template and adapted it accordingly.

I did run into the issue that `zip` cli is not cross-platform, so I had to add `archiver`, and wrote a quick script to do the same that was being done with the other tools.

I also added a section that will run if it's not in CI that will reuse some packages already in the dependency tree to provide progress bars when zipping files to help with debugging it.

See https://github.com/josh-hemphill/devtools/releases/tag/v6.4.0 for what the successful release with artifacts looks like.


josh-hemphill:
Okay, I think it's good. I had to disable webpack-terser parallelism to pass CircleCI without it running out of memory, which also seems to be what's effecting other PRs.

Akryum:
Thanks! :+1:

Issue #1532, opened 2021-08-11
Shortcut (Ctrl+F) for filter component by name input box doesn't work


dkjain:
### Version
5.3.4

### Browser and OS info
Chrome 92 / Windows 7

### Steps to reproduce
In Vue devtools panel, `Ctrl   F` shortcut is defined for filter component by name input box, (see image), but it seems doesn't work anymore, Its overridden by "Search box for text within the current panel". 

![vue-devtools-components-list-view](https://user-images.githubusercontent.com/10743577/129031839-cc4e6b03-2165-40e2-887f-3a07501f093f.png)

### What is expected?
Using `Ctrl   F` shortcut defined for _filter component by name_ input box doesn't work
![vue-devtools-components-list-view](https://user-images.githubusercontent.com/10743577/129031817-084cb3d9-6284-44ed-9189-a18d1e23b8dd.png)

### What is actually happening?
Using `Ctrl   F` shortcut in vue devtools focuses _Search box for text within the current panel"_.

---
Request is to change the keyboard shortcut to be more unique something like `Ctrl Shift F` or `Ctrl Shift K` since `Ctrl   F` is almost a standard keyboard shortcut across window applications.

Even, better would be the _ability to specify custom shortcuts_ for this or all shortcuts of vue devtools extension.

dkjain:
Also, there is an Chrome extension API specifically for this purpose rather than having to bake in the extension itself, 

![devtools-chrome-commands-api](https://user-images.githubusercontent.com/10743577/129033518-55f9b5a5-f586-48b7-9dff-ed45600cea54.png)

[chrome.commands API](https://developer.chrome.com/docs/extensions/reference/commands/)

This way the extension can have something like this:

![ublock-origin-keyboard-shortcuts-editor-in-chrome](https://user-images.githubusercontent.com/10743577/129033633-c17e7937-6cd7-4e40-9ade-0cb60556caaa.png)

konpeki622:
Seems like that the tooltip about `Ctrl + F` shortcut has been removed on v6. Not sure whether we would change the keyboard shortcut or not.

Issue #1281, opened 2020-09-18
Add keyboard shortcut for "Filter Inspected data/state"


dkjain:
### What problem does this feature solve?
I'd like to have keyboard shortcut to quickly gain focus to "Filter inspected data/state" input box. Additionally I would prefer if the input box gaining focus when the Component or VueX tab is switched to. This greatly improves UX as the inputs are the most used by developers IMHO.

Thanks

### What does the proposed API look like?
Keyboard shortcut to quickly gain focus to "Filter inspected data/state" input boxes.

PR #1598, opened 2021-11-02
feat: modify shortcut and add tooltips

konpeki622:
According to #1532 and #1281 , I try to make some modifications. Shortcut of filter components was changed from `Ctrl + F` to `Alt + F`. After select a component, we can press `Alt + D` to filter state. The same time related tooltip was added.

### DIFF
--- a/packages/app-frontend/src/features/components/ComponentsInspector.vue
+++ b/packages/app-frontend/src/features/components/ComponentsInspector.vue
@@ -42,7 +42,7 @@ export default defineComponent({
     } = useComponentPick()
 
     onKeyDown(event => {
-      if (event.key === 'f' && event.ctrlKey) {
+      if (event.key === 'f' && event.altKey) {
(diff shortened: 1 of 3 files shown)


arpit-uno:
@Akryum These new shortcuts `Alt + F` and `Alt + D` overlaps with Chrome's shortcuts mean, if Filter State input box is focused (`Alt + D`) and getting focus Filter component input (`Alt + F`), chromes three vertical dot menu (towards top-right) button/trigger steals focus. And same when filter component input is focused (`Alt + F`) and now trying to focus filter state input (`Alt + D`), chrome's omnibar gets focused.

Pls test it.

dkjain:
@Akryum Is it possible to provide a shortcuts editor for users to set shortcut to their preferences? Irrespective if this is possible or not, the current shortcuts overlaps with Chromes shortcuts and will need to be changed

PR #1611, opened 2021-11-18
refactor: node_modules/vuex/dist/vuex.esm-bundler.js import { setupDe…

panhezeng:
…vtoolsPlugin } from '@vue/devtools-api';

          The production model is packaged into the vendor.
          The proxy that vue 3 relies on supports at least Chrome 49, Spread syntax in object literals supports at least Chrome 60.


bastiW:
+1 for this. 

This seems to fix bugs in older browsers of the router https://github.com/vuejs/vue-router-next/issues/1207

PR #1620, opened 2021-11-26
refactor: inherit inactive state

### LINKED ISSUE #1615: Active child components disappered when `KeepAlive` is selected
### Version
6.0.0-beta.20

### Browser and OS info
Chrome 93.0.4577.63, Windows 10.

### Steps to reproduce
1. init a vue3 with vue-router4 project.
2. create components and config router.
3. use `<router-view>` and `<keep-alive>`.
4. select `KeepAlive` in component tree of devtools.

Reproduction: [vue-devtools-debug](https://github.com/konpeki622/vue-devtools-debug/tree/main/vue3-test)
It does not occur in vue2 project. The same problem before version 6.0.0-beta.18.

### What is expected?
It works well.

Active child components disappeared in component tree:
vue3keepalive

konpeki622:
Previously, the cached components were pushed into `treeNode.children` and they replaced the active components, which caused #1615 . So I added `cache` to `ComponentTreeNode` to fix it. At the same time I removed some recursions to improve performance.

### DIFF
--- a/packages/api/src/api/component.ts
+++ b/packages/api/src/api/component.ts
@@ -12,6 +12,7 @@ export interface ComponentTreeNode {
   isFragment: boolean
   hasChildren: boolean
   children: ComponentTreeNode[]
+  cache?: ComponentTreeNode[]
   domOrder?: number[]
   consoleId?: string
   isRouterView?: boolean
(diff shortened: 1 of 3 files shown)


Akryum:
Hi :green_heart: ! Thank for your PR, but the issue you linked was already solved by https://github.com/vuejs/devtools/commit/54b2be8394948a4e284218facb9315be3eaca97e

Why is this PR needed?

konpeki622:
Hi, @Akryum .

I think we called `capture` too many times. It will works with poor performance when the child component of keepalive has deep nesting. To avoid that, I modified the method of setting inactive state. What do you think about that?

konpeki622:
I will close this PR because this changing may increase the complexity of filter. 😮

Issue #1587, opened 2021-10-19
States lost after HMR


konpeki622:
### Version
6.0.0 beta 19

### Browser and OS info
Chrome 93.0.4577.63, Windows 10

### Steps to reproduce
1. init a project with vue 3.x
2. get components by filter and select one
3. edit some variables in this component and trigger HMR

### What is expected?
New values were shown on the devtools panel.

### What is actually happening?
States lost.

Before:
![before](https://user-images.githubusercontent.com/28253544/137841255-7d5ff792-0471-4c05-afaa-bd60ea099d40.PNG)

After HMR:
![after](https://user-images.githubusercontent.com/28253544/137841276-5e3c59a3-d2c7-4261-84a3-8d6c2792eca1.PNG)

konpeki622:
The component id were changed when refreshing but the selected id stayed. I think the issue might be caused by it.

nsrosenqvist:
Experiencing this issue too with Vite v5 HMR.

nsrosenqvist:
I pulled down the development environment and am currently investigating this issue through the chrome dev environment. It appears that it is working fine as long as there are more than one component in a component tree, but when there's only a single one, and I try to select it in the component tree after a HMR update, Vue Devtools seems to get confused and `app-1:root` gets replaced in the instance map by `app-1:1`. Therefore, it doesn't find the app record for `app-1:root` any longer. @konpeki622 @Akryum any idea where I should start looking?

x0rsw1tch:
@nsrosenqvist I've been trying to figure out why this was occurring on some components I've been working on. `vue-devtools` in the browser just seems to "forget" where the app is, and refuses to update after an HMR change, have to reload the page to fix.

Per your findings, I tried attaching a simple component to one of the troubled `.vue` files I'm working on, which previously had no child components on it... With that dummy component attached to the root instance, it continues to function within vue-devtools after an HMR update now, strange! It also seems to work fine by adding one of the built-in components. I tacked on a `<transition />` to the end of the template, and that also seems to fix it.

FWIW, when it's not working, and have debugging info enabled, this shows up in the log:

```
Instance uid=cloudagentsworldmappoints:root not found
getComponentInstance @ backend.js:1596
sendSelectedComponentData @ backend.js:1497
(anonymous) @ backend.js:2315
wrappedListener @ backend.js:7245
emit @ backend.js:10901
_emit @ backend.js:7293
(anonymous) @ backend.js:7231
[...]

PR #1625, opened 2021-12-02
fix: refresh component tree

konpeki622:
Fix and close the issue #1587 .


konpeki622:
To fix #1633 , old deactivated instance should be removed when new instance was mounted. It also refers to the lifecycle hooks of keepalive components which I mentioned in #1624 .

nsrosenqvist:
@konpeki622 #1587 appears to not have been fixed by this, is there anything I can help with to resolve this issue?

Issue #1543, opened 2021-08-19
Collapse all properties


debug87:
Why are all (or some) properties always expanded? It is incredibly annoying when everything is expanded every time and you have to collapse it to get a clear view.
All properties (data, props, etc.) should always be collapsed.

piotrmachacki:
Please fix this because it's currently hard to use when everything is expanded.

konpeki622:
For collapse all properties, press `ctrl` or `meta` on keyboard while clicking the toggle. For expand press `shift` instead.

Akryum:
Duplicate of #1391

PR #1635, opened 2021-12-13
feat: collapse all properties defaultly

konpeki622:
Related to issue #1543 .


konpeki622:
It will be used like this:
result

PR #1640, opened 2021-12-20
feat: overwrite stringify for copying

### LINKED ISSUE #1002: Copy strings to clipboard without quotes
I find it really annoying how copied strings from the data inspector always include double-quotes. Should that really be the default behaviour?

At the moment, the clipboard value comes from here:
https://github.com/vuejs/vue-devtools/blob/21552134f022be2b8976d205a5c6a23f5ef77fc7/src/devtools/components/DataField.vue#L474
https://github.com/vuejs/vue-devtools/blob/591c1c235f43ab201d4ca5c1e2ab00be4ca3b46f/src/util.js#L616
https://github.com/vuejs/vue-devtools/blob/591c1c235f43ab201d4ca5c1e2ab00be4ca3b46f/src/util.js#L115

There would be many ways to change that behaviour, but first I would like to know if you agree.

konpeki622:
Closes #1002 .

You can test it on [codesandbox](https://codesandbox.io/embed/kind-joana-mu0fw?fontsize=14&hidenavigation=1&theme=dark).

### DIFF
--- a/packages/shared-utils/src/transfer.ts
+++ b/packages/shared-utils/src/transfer.ts
@@ -58,49 +58,69 @@ function decode (list, reviver) {
   }
 }
 
+function serializeArray (data: any, replacer: (this: any, key: string, value: any) => any = null, space: number = null, lastSpace: number = null) {
+  const objKeys = Object.keys(data)
+  if (objKeys.length === 0) {
+    return '[]'
+  }
+  let result = '['
+  const newSpace = lastSpace + space
+  objKeys.forEach(key => {
+    const valueString = stringifyWithReplacer(data[key], replacer, space, newSpace)
+    result += space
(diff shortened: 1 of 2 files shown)


konpeki622:
I modified the rule. The quote is remained if it was string value in object. And like `JSON.stringify`, the `symbol`, `function` and `undefined` will be ignored when there is no replacer.

Copying path and value:
image

You can test it in [codesandbox](https://codesandbox.io/embed/kind-joana-mu0fw?fontsize=14&hidenavigation=1&theme=dark).

Akryum:
This looks a bit complicated and there are probably a lot of cases not covered as serializing JS data into JS code is a complex subject. Only checking if the root value is a primitive would be more than enough.

Akryum:
See 7a3e42f569fb61cea7c41cb4a4e92d34d01ddadc

PR #1657, opened 2022-01-06
fix: support vue2 app debugging in micro-app

### LINKED ISSUE #1656: Cannot register multiple vue2 apps in devtools with micro-app
### Version
6.0.0-beta.21

### Browser and OS info
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.110 Safari/537.36 Edg/96.0.1054.62

### Steps to reproduce
Here's the repo: https://github.com/DevilTea/reproduce-devtools-micro-app-issue

### What is expected?
Expected to debugging multiple vue2 app with microfrontend framework [micro-app](https://github.com/micro-zoe/micro-app)

The devtools only shows single vue2 App

DevilTea:
closes #1656

### DIFF
--- a/packages/app-backend-core/src/legacy/scan.ts
+++ b/packages/app-backend-core/src/legacy/scan.ts
@@ -1,4 +1,5 @@
 import { isBrowser, target } from '@vue-devtools/shared-utils'
+import { getPageConfig } from '../page-config'
 
 const rootInstances = []
 
@@ -56,6 +57,17 @@ export function scan () {
     for (const iframe of iframes) {
       try {
         walkDocument(iframe.contentDocument)
+      } catch (e) {
+        // Ignore
+      }
+    }
+
+    // Scan for Vue instances in the customTarget elements
+    const { customVue2ScanSelector } = getPageConfig() as { customVue2ScanSelector: string }
+    const customTargets = customVue2ScanSelector ? document.querySelectorAll(customVue2ScanSelector) : []
+    for (const customTarget of customTargets) {
+      try {
+        walkDocument(customTarget)
       } catch (e) {
         // Ignore
       }
--- a/packages/app-backend-core/src/legacy/scan.ts
+++ b/packages/app-backend-core/src/legacy/scan.ts
@@ -63,7 +63,7 @@ export function scan () {
     }
 
     // Scan for Vue instances in the customTarget elements
-    const { customVue2ScanSelector } = getPageConfig() as { customVue2ScanSelector: string }
+    const { customVue2ScanSelector } = getPageConfig()
     const customTargets = customVue2ScanSelector ? document.querySelectorAll(customVue2ScanSelector) : []
     for (const customTarget of customTargets) {
       try {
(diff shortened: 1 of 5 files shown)


packages/app-backend-core/src/legacy/scan.ts
```diff
@@ -60,6 +60,16 @@ export function scan () {
         // Ignore
       }
     }
+
+    // Scan for Vue instances in the <micro-app> elements
+    const microApps = document.querySelectorAll('micro-app')
```
Akryum, line 65: We should probably let users configure this selector in case they use something else than micro-app?

DevilTea: Oh, that sounds really flexible~
Maybe I could try to put it into GlobalSettings!

Akryum: Maybe it would be better on `window.VUE_DEVTOOLS_CONFIG` since it's specific to each app. Example:
https://devtools.vuejs.org/guide/open-in-editor.html#customize-request

You can probably use `getPageConfig` from `page-config.ts` in backend core

DevilTea: @Akryum I've added it into GlobalSettings. It looks like following images.

image

image

DevilTea: > Maybe it would be better on `window.VUE_DEVTOOLS_CONFIG` since it's specific to each app. Example: https://devtools.vuejs.org/guide/open-in-editor.html#customize-request
> 
> You can probably use `getPageConfig` from `page-config.ts` in backend core

Okay I would try it~ Thank you for your advice!

DevilTea: I have change to use `getPageConfig ` to get the config and add a section in docs to explain the config, it works great to me!

DevilTea: > @Akryum I've added it into GlobalSettings. It looks like following images.
> 
> image
> 
> image

The part of this is removed.

packages/app-backend-core/src/legacy/scan.ts
```diff
@@ -60,6 +61,17 @@ export function scan () {
         // Ignore
       }
     }
+
+    // Scan for Vue instances in the customTarget elements
+    const { customVue2ScanSelector } = getPageConfig() as { customVue2ScanSelector: string }
```
Akryum, line 66: Could you add the field to the `PageConfig` interface instead of casting?
image

Akryum (changes requested):

packages/app-backend-core/src/page-config.ts
```diff
@@ -3,6 +3,7 @@ import { target, SharedData } from '@vue-devtools/shared-utils'
 export interface PageConfig {
   openInEditorHost?: string
   defaultSelectedAppId?: string
+  customVue2ScanSelector: string
```
Akryum, line 6:
```suggestion
  customVue2ScanSelector?: string
```

DevilTea: Ok, I've updated!

DevilTea:
Hello @Akryum , is there anything else I need to do?

DevilTea:
Sorry for bothering again, @Akryum would this pr be merged in next release?
Is there anything else I need to do to make this pr be able to be merged?
Thank you~

PR #1690, opened 2022-02-09
fix(vue2): Use the super constructor when the instance is a VueComponent

### LINKED ISSUE #1687: Vue2 fails when using a VueComponent instead of Vue as root
### Version
6.0.2

### Reproduction link
[codesandbox.io](https://codesandbox.io/s/gracious-hooks-oicom)

### Steps to reproduce
Use Vue 2
Create a component using `Vue.extend()`
Create an instance of the component using `new` and mount it on the page.
Open Vue dev tools.

### What is expected?
The component can be inspected using Vue dev tools

An exception is thrown
```javascript
Uncaught (in promise) TypeError: options.version is undefined
    registerAppJob backend.js:949
...
```

---
The root cause seems to be in this function.
https://github.com/vuejs/devtools/blob/b3dd744c6ec3a4eeb200b8003e0ff74b9e78abfe/packages/app-backend-core/src/app.ts#L235

In some cases (e.g. `Vue.extend`) `app.constructor` points to `VueComponent` instead of `Vue`. This class does not have a version property.

0Rick0:
Fixes #1687

### DIFF
--- a/packages/app-backend-core/src/app.ts
+++ b/packages/app-backend-core/src/app.ts
@@ -222,6 +222,16 @@ export async function removeApp (app: App, ctx: BackendContext) {
 }
 
 // eslint-disable-next-line camelcase
+function _legacy_getVueFromApp (app) {
+  if (app.constructor.name === 'VueComponent') {
+    // When Vue.extend is used the component is an instance of VueComponent instead of Vue.
+    // VueComponent has a property super which points to the original Vue constructor
+    return app.constructor.super
+  }
+  return app.constructor
+}
+
+// eslint-disable-next-line camelcase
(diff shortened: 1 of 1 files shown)


packages/app-backend-core/src/app.ts
```diff
@@ -221,6 +221,14 @@ export async function removeApp (app: App, ctx: BackendContext) {
   }
 }
 
+// eslint-disable-next-line camelcase
+function _legacy_getVueFromApp (app) {
```
Akryum, line 225: Could you add a comment explaining why we need to do this? Thanks!

Akryum (changes requested):

Akryum (approved):

masstez (approved):

PR #1719, opened 2022-02-11
fix: generated app id not deduped

soultice:
This fixes a bug where more than two components with the same name could not be held apart by the backend. It is looking for `defaultId:count`while counting references but then names the reference `defaultId_count` this resulted in a maximum of two components with the same name being recognized (that is clickable) in the devtools.

### DIFF
--- a/packages/app-backend-core/src/app.ts
+++ b/packages/app-backend-core/src/app.ts
@@ -139,7 +139,7 @@ export function getAppRecordId (app, defaultId?: string): string {
 
   if (defaultId && appIds.has(id)) {
     let count = 1
-    while (appIds.has(`${defaultId}:${count}`)) {
+    while (appIds.has(`${defaultId}_${count}`)) {
       count++
     }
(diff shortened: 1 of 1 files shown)


packages/app-backend-core/src/app.ts
```diff
@@ -142,7 +142,7 @@ export function getAppRecordId (app, defaultId?: string): string {
     while (appIds.has(`${defaultId}:${count}`)) {
```
Akryum, line 142: It should be changed here instead

Akryum (changes requested):
Good catch

soultice:
Hope I got you right and you want `_` for the separator?

Akryum:
Yes because now `:` should only used to separate the app id and the component id

Akryum (approved):

soultice:
Thanks for the merge, greatly appreciated :+1:

PR #1783, opened 2022-03-10
Fix：fix find components input overflow

Starlogzzz:
when the width of the Components is small, the input of Find components will overflow
devtools


Akryum:
Hi! Thank you for the time you took to open your PR :green_heart: ! However I fixed the issue you describe [here](https://github.com/vuejs/devtools/commit/94fe74e86b903624985bc6aa0da56b117612b9cf) differently and also for the custom inspectors - so I'm gonna close yours.

Issue #2046, opened 2023-04-07
devtools 不识别camelCase 形式命名的事件


dfjfj:
### Vue devtools version

6.5.0

### Link to minimal reproduction

[...]

PR #2051, opened 2023-04-13
fix(vue3): Fix the declared camelCase event cannot be recognized

Azurewarth0920:
### Description

fix: #2046 
fix: #2063
fix: #2060

Fix the declared camelCase event that cannot be recognized.

the declaredEmits are not normalized to kebab case, which results in the mismatching with the declared events on the parent component side.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/app-backend-vue3/src/components/data.ts
+++ b/packages/app-backend-vue3/src/components/data.ts
@@ -371,21 +371,24 @@ function processRefs (instance) {
 function processEventListeners (instance) {
   const emitsDefinition = instance.type.emits
   const declaredEmits = Array.isArray(emitsDefinition) ? emitsDefinition : Object.keys(emitsDefinition ?? {})
-  const normalizedDeclaredEmits = declaredEmits.map(key => kebabize(key))
+  const declaredEmitsMap = declaredEmits.reduce((emitsMap, key) => {
+    emitsMap[kebabize(key)] = key
+    return emitsMap
+  }, {})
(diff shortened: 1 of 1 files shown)


packages/app-backend-vue3/src/components/data.ts
```diff
@@ -371,13 +371,14 @@ function processRefs (instance) {
 function processEventListeners (instance) {
   const emitsDefinition = instance.type.emits
   const declaredEmits = Array.isArray(emitsDefinition) ? emitsDefinition : Object.keys(emitsDefinition ?? {})
+  const normalizedDeclaredEmits = declaredEmits.map(key => kebabize(key))
   const keys = Object.keys(instance.vnode.props ?? {})
   const result = []
   for (const key of keys) {
     const [prefix, ...eventNameParts] = key.split(/(?=[A-Z])/)
     if (prefix === 'on') {
       const eventName = eventNameParts.join('-').toLowerCase()
```
Azurewarth0920, line 380: This line makes the eventName kebabized.

But the variable [declaredEmits](https://github.com/vuejs/devtools/pull/2051/files#diff-554ccf9aaff1681ee888264aad2cc1a53b7db9271dcf287c655ee6c79b4ea866R373) is a mixed case array, which will cause a mismatch.

lbineau:
Sale issue, it is quite missleading to have warnings from Vue Devtools if we use camelCase instead of kebab-case for events.

PR #2082, opened 2023-06-29
fix(component): filtering components of third-party libraries to script(other)

KaevN:
### Description

Problem: if you use third-party ui libraries, the instance of the used components gets into the "setup" block and is mixed with the instance of your component, which is very annoying.
bad

Solution: Added check for component and redirect components to "script(other)" block
good

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


KaevN:
@Akryum Hello. will it be possible to add this to the next release? instance of components of third-party libraries is very disturbing in the setup block. The screenshots show an example of when you use Ant Design Vue components. Or write if you have any comments or problems with this change

PR #2119, opened 2023-10-19
refactor(frontend): Migrate the frontend application from Vue.js 2 to 3

Azurewarth0920:
### Description

EOF of Vue.js 2 is at the end of this year, so I tried to migrate to Vue 3.

- Updated library and rewrote the frontend application with Vue 3 syntax.
- The project uses the UI library [@vue/ui](https://github.com/vuejs/ui) and seems not under maintenance, moves the components here and rewrites them with composition API.
- The tests are broken so I manually packaged it and tested it with my project, and it seems to work well.

### Additional context

The PR has huge diffs and may be hard to review, and I try to separate each migration item into commits,
so that reviewing commit by commit is easier to read. Can you take a look at this? @Akryum 

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


Akryum:
👀

Akryum:
Thank you very much! :pray:

packages/app-frontend/src/features/components/ui/components/icons.ts
```diff
@@ -0,0 +1,34 @@
...
+  /materialicons\/24px\.svg$/,
+)
+
+export default {
+  install () {
+    const sprites = ['']
+    let spriteIndex = 0
+    // Load all the SVG symbols
+    icons.keys().forEach((key, index) => {
+      let result = icons(key)
```
sfe-efficy, line 13: Hi!

This code is return a base64 encoded content with the string "module.exports = ....", I guess is normal because we are using require.context.

So, at the end is reporting errors all code at line 15 because the regular expresions will return a null (then cannot be destruct from null).

I tested with a code like:

```
      const base64Prefix = 'data:image/svg+xml;base64,'
      if (result.startsWith(base64Prefix)) {
        const positionPrefix = result.indexOf(base64Prefix) + base64Prefix.length
        const base64Content = result.substring(positionPrefix)
        result = atob(base64Content)
        const module: { exports?: string } = {}
        eval(result)
        result = module.exports
      }
```

before using the regular expression.

